PR #32, opened 2026-07-07
⚡ Bolt: [performance improvement] Non-blocking concurrent file search

giauphan:
💡 **What:** The optimization replaces the synchronous `fs.readFileSync` in a loop with an asynchronous worker pool using `fs.promises.readFile`.

🎯 **Why:** The synchronous file reading loop inside the `code_search` tool was blocking the Node.js event loop for the entire duration of the search (up to several hundreds of milliseconds for large filesystems). This prevented other network requests and background tasks from being processed.

📊 **Measured Improvement:**
Before (Sync): Execution time ~363ms. Max Event Loop Delay/Blocking: ~363ms.
After (Async Worker Pool 20): Execution time ~693ms. Max Event Loop Delay: ~5ms.

Note on execution time: The overall execution time increases for local search because asynchronous I/O scheduling in node introduces overhead. However, it's a huge win for server scalability as it frees up the event loop for hundreds of milliseconds that were previously completely blocked.

---
*PR created automatically by Jules for task [6612948511315592319](https://jules.google.com/task/6612948511315592319) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,0 +1,6 @@
+## Bolt Journal (Performance Optimizations)
+
+**Goal:** Ensure the codebase is fast, efficient, and scalable.
+
+### Learning
+- 2023-11-20: Optimizing synchronous file I/O operations (`fs.readFileSync`) to asynchronous ones (`fs.promises.readFile`) in a tight loop is critical for unblocking the event loop, especially when dealing with hundreds or thousands of files. Using an asynchronous worker pool pattern strikes the perfect balance between high concurrency, early exit capability (to stop processing once a `maxRes` threshold is reached), and event loop responsiveness.
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1551,6 +1551,7 @@ export function registerTools(server: McpServer) {
       const results: Array<{ file: string; line: number; content: string; contextBefore: string[]; contextAfter: string[] }> = [];
       const concurrency = 20;
       let currentIndex = 0;
+      const projectDir = loaded.projectDir;
 
       async function worker() {
         while (currentIndex < allFiles.length && results.length < maxRes) {
@@ -1563,7 +1564,7 @@ export function registerTools(server: McpServer) {
               if (results.length >= maxRes) break;
               if (lines[i].toLowerCase().includes(q)) {
                 results.push({
-                  file: path.relative(loaded!.projectDir, filePath),
+                  file: path.relative(projectDir, filePath),
                   line: i + 1,
                   content: lines[i].trim(),
                   contextBefore: lines.slice(Math.max(0, i - ctx), i).map(l => l.trim()).filter(Boolean),
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1549,25 +1549,43 @@ export function registerTools(server: McpServer) {
       } catch { /* fallback */ }
 
       const results: Array<{ file: string; line: number; content: string; contextBefore: string[]; contextAfter: string[] }> = [];
-      for (const filePath of allFiles) {
-        if (results.length >= maxRes) break;
-        try {
-          const content = fs.readFileSync(filePath, "utf-8");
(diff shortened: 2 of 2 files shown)


giauphan:
cc @jules

This is a clean, correct refactor from synchronous to async file I/O with a worker pool. No real issues.

The concurrent worker pattern is safe — Node.js's single-threaded event loop guarantees that `results.push()` and `currentIndex++` are serialized because there's no `await` between any of those operations within a single file's processing turn. Workers only yield at `fs.promises.readFile`, so there's no data race on the shared `results` array.

The sorting at the end correctly restores deterministic ordering. The `results.slice(0, maxRes)` in the return handles the edge case where we might push one extra result beyond `maxRes` between workers.

**Only minor observations (none blocking):**

- At most ~20 extra files can be read unnecessarily due to workers already having claimed their file before `results.length >= maxRes` propagates. Negligible.
- `loaded!` uses a non-null assertion that wasn't needed before — same type safety guarantee, just satisfying a narrower narrowing in the async context.
- The `.jules/bolt.md` journal is documentation-only and has no effect on behavior.

Nothing to flag on security, performance, or maintainability. Ship it.

---

giauphan:
cc @jules

**Correctness**

- **Shared `currentIndex` with race:** `currentIndex++` races between workers because it's a read-modify-write. Two workers can read the same index, then both increment — one file gets processed twice, another is skipped. Fix: use an atomic counter or have a single producer dispatching work. 

- **`results.length` bounds checking isn't safe:** multiple workers can all pass `results.length < maxRes`, then all `push` results, exceeding `maxRes`. The `slice(0, maxRes)` in the return mitigates this, but you'll still do wasteful I/O once the limit is hit. The same issue applies to the early-exit `return` in the inner loop.

**Security**

No issues introduced. Same paths, same file operations, same `try/catch { /* skip */ }` pattern.

**Performance**

Async I/O is a clear improvement over synchronous blocking reads in a loop. The concurrency of 20 is reasonable for file reads. However, once `maxRes` is reached, in-flight `readFile` calls still complete before their workers return — wasted I/O. A shared abort signal would prevent this.

**Code Quality / Maintainability**

- The `.jules/bolt.md` file is a journal entry — no code impact, fine as documentation.
- The worker pool pattern is appropriate.
- **Missing variable:** `maxResults` on line `results.slice(0, maxRes)` — assuming this is consistent with the function parameter, but double-check the variable name matches.

**Verdict:** The approach is sound and the async conversion is good. **Fix the `currentIndex` race** (the one real bug) — either use a lock-free shared index via `Atomics.add` on a `SharedArrayBuffer`, or assign work upfront in chunks instead of racing on `currentIndex++`. Everything else is minor enough to ship after that fix.

---

giauphan:
cc @jules

This PR looks clean. Let me walk through the key points:

**Correctness**: The async worker pool (concurrency=20) replaces synchronous `readFileSync` with `fs.promises.readFile`. The `currentIndex` increment is safe — it happens synchronously before the `await`, so no two workers get the same file. The `results.length >= maxRes` guard after `readFile` is a best-effort check — workers that were already mid-read when the threshold was hit will still push their results, potentially exceeding `maxRes`. This is harmless because the response uses `results.slice(0, maxRes)`.

**Security**: No issues. Same file ops, same path sanitization via `path.relative`.

**Performance**: This is a real improvement. The synchronous loop blocked the event loop for the entire search; now 20 files are read concurrently with proper yielding. The final `sort()` is O(n log n) on results (small) and ensures deterministic output.

**Code quality / Maintainability**: No dead code, no duplication. The early `return` after `readFile` when `results.length >= maxRes` is a minor optimization quirk — it exits one worker early while others finish their in-flight file — but it doesn't affect correctness. The `.jules/bolt.md` journal addition is just documentation, no concerns.

The only thing worth a mention is that with 20 concurrent workers you could process up to 19 extra files past the `maxRes` threshold (all workers read their file before checking the guard after the `await`). This is negligible given the file I/O overhead but worth noting if `maxRes` is small and files are large. Not a blocker.

---

PR #34, opened 2026-07-09
🛡️ Sentinel: [CRITICAL] Fix API Key Injection Vulnerability

giauphan:
🚨 Severity: CRITICAL
💡 Vulnerability: Code injection/RCE vulnerability in setup_second_brain tool. Input apiKey from the user was directly embedded into a python script and YAML config without validation.
🎯 Impact: This could allow an attacker to achieve arbitrary code execution via Python strings or YAML configuration injection.
🔧 Fix: Added a regex validation check for the API key input to ensure it contains only expected characters (alphanumeric, dots, dashes, underscores).
✅ Verification: Ran `npm run test` and all tests passed. Validated via `git diff` that the correct regex check is in place.

---
*PR created automatically by Jules for task [9486243253122884246](https://jules.google.com/task/9486243253122884246) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -6,3 +6,7 @@
 **Vulnerability:** dreamingService API calls did not include the apiKey in the query parameter despite the tests demanding it.
 **Learning:** The implementation for appending apiKey query parameter was missing.
 **Prevention:** Ensure tests comprehensively cover all authentication logic and that API parameters are consistently added.
+## 2024-07-09 - [Code Injection via Unvalidated Input]
+**Vulnerability:** Code injection/RCE vulnerability in setup_second_brain tool
+**Learning:** Input apiKey from the user was directly embedded into a python script without validation.
+**Prevention:** Validate input format (using a restrictive regex) before embedding it into configuration files or scripts.
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1976,11 +1976,10 @@ export function registerTools(server: McpServer) {
       const auth = await checkAuth();
       await logActivity(auth, "setup_second_brain", { client, autoPlugin });
 
-      let key = apiKey || process.env.CODEATLAS_API_KEY;
+      const key = (apiKey || process.env.CODEATLAS_API_KEY)?.trim();
       if (!key) return { content: [{ type: "text" as const, text: JSON.stringify({
         success: false, error: "CODEATLAS_API_KEY not set. Provide apiKey parameter or set env var."
       }, null, 2) }] };
-      key = key.trim();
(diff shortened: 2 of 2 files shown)


giauphan:
cc @jules

**Validation placement is correct** — it gates access before the key reaches the template literal in `mcpEntry` (line ~1989), so the injection vector is properly blocked.

**Regex correctness**: `/^[a-zA-Z0-9_\-.]+$/` — the escaped `\-` is a literal hyphen, `.` is literal inside `[]`, and `+` rejects empty strings. All looks fine for typical API key formats.

One minor nit worth flagging:

- **Trailing/leading whitespace**: If a user passes `" sk-abc123 "` (accidental space), the regex rejects it outright rather than trimming. A `.trim()` before validation would be more user-friendly. Not a security issue (the rejection is safe), just a UX polish item.

Otherwise this is clean.

---

giauphan:
cc @jules

This is a well-scoped security fix. Here's my analysis:

**No correctness, performance, or maintainability issues.** The validation is placed at the right point, the error message is clear, and the regex check is O(n) on a short string — negligible.

**The only issue I'd flag:**

The `const` → `let` change is unnecessary. `key` is only reassigned once (`.trim()`), so this can be expressed as a single `const`:

```ts
const key = (apiKey || process.env.CODEATLAS_API_KEY)?.trim();
```

Minor, but keeping `const` signals the value won't be mutated after initialization, which is cleaner.

**Everything else looks good.** The regex `^[a-zA-Z0-9_\-.]+$` correctly disallows `$`, backticks, quotes, semicolons, newlines, and other injection-relevant characters that could break out of the YAML double-quoted string context. The sentinel entry documents the learning for future reference.

---

giauphan:
cc @jules

This is a clean, focused security fix. I only have one minor observation.

**Security** — The regex `^[a-zA-Z0-9_\-.]+$` effectively prevents YAML injection. The key was previously interpolated raw via template literal, then placed inside double-quoted YAML. Without validation, a key containing `"` or `\n` could break the config structure. The restrictive character whitelist eliminates that. The `.trim()` before validation is also correct — catches trailing newlines that might slip in from env vars.

**Correctness** — The escaped hyphen `\-` inside the character class is valid JavaScript regex (treated as a literal `-`). The escaped dot `\.` is unnecessary inside `[]` but harmless. One edge case: the regex requires **one or more** characters (`+`), so an empty string after trim is caught by the `!key` guard above rather than this regex check — that's fine, the first guard handles it.

**The rest** (performance, code quality, maintainability) — No issues. The change is minimal, well-placed, and the sentinel entry documents the lesson for future contributors. Good to merge.

---

PR #36, opened 2026-07-10
⚡ Bolt: [parser performance improvement]

giauphan:
💡 **What**: Refactored `buildAnalysisResult` and `generateAIInsights` in `src/analyzer/parser.ts` to use single-pass loops instead of multiple iterations, and pre-computed relationship counts in a Map to eliminate an O(N * L) nested loop.
🎯 **Why**: The previous implementation performed multiple O(N) passes through the nodes list to count entity types, creating unnecessary intermediate arrays. More critically, checking for God Objects inside `generateAIInsights` had an O(N*L) time complexity because it filtered `graph.links` for every single module node, which slows down analysis significantly on large codebases.
📊 **Impact**: Reduces time complexity of insights generation from O(N * L) to O(N + L). Reduces the four O(N) iterations over nodes in `buildAnalysisResult` down to a single O(N) pass, and completely avoids intermediate array allocations created by `Array.from().filter()`.
🔬 **Measurement**: Time complexity reduced. Can be measured by testing analysis runtime on large projects containing 5,000+ files.

---
*PR created automatically by Jules for task [5667617937203090144](https://jules.google.com/task/5667617937203090144) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,9 +1,3 @@
-## Bolt Journal (Performance Optimizations)
-
-**Goal:** Ensure the codebase is fast, efficient, and scalable.
-
-### Learning
-- 2023-11-20: Optimizing synchronous file I/O operations (`fs.readFileSync`) to asynchronous ones (`fs.promises.readFile`) in a tight loop is critical for unblocking the event loop, especially when dealing with hundreds or thousands of files. Using an asynchronous worker pool pattern strikes the perfect balance between high concurrency, early exit capability (to stop processing once a `maxRes` threshold is reached), and event loop responsiveness.
-### 2024-05-24 - O(N×L) Node Degree Calculation Bottleneck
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

This looks clean. Here's my review:

**Correctness:** Both optimizations are semantically equivalent. The `modulesWithManyFunctions` change correctly builds an array of `node.id` strings directly instead of mapping later — no bug. The link filtering condition (`link.type === 'import' && link.target.startsWith('function')`) is preserved identically in the pre-computation Map.

**Security:** Nothing to flag — pure computation, no user input, I/O, or secrets.

**Performance:** Achieves the stated O(N*L) → O(N+L) transformation correctly. The `buildAnalysisResult` change replaces 5 separate passes (4 × `Array.from().filter()`, 1 × `filter()`) with 2 single-pass loops. The `generateAIInsights` change replaces a nested filter-inside-loop with a Map pre-computation. Both are correct.

**Code quality / Maintainability:** No issues. The `if-else if` chain for the 4 node types is simple and readable — a lookup table would be over-engineering for 4 cases. The variable naming is clear.

One minor observation (not a blocker): the `generateAIInsights` pre-computation iterates `graph.links`, while `buildAnalysisResult` iterates `this.links`. If these two link arrays are the same object, this is fine — but if they could diverge, there's a potential inconsistency. That said, the old code also used `graph.links` in the same function, so this isn't a regression.

**Verdict:** Ready to merge. No bugs, no regressions.

---

PR #50, opened 2026-07-14
🧹 [Code Health] Extract path resolution to utility functions

giauphan:
🎯 **What:** Extracted repetitive `path.join(os.homedir(), ...)` logic into shared utility functions in `src/utils/pathUtils.ts`. Updated references in `src/presentation/mcpServer.ts`, `src/cli/commands.ts`, and `src/services/projectService.ts`.
💡 **Why:** Reduces duplicate code, improves maintainability, and makes path handling consistent across the application.
✅ **Verification:** Compiled code successfully, ran `npm run test` with no regressions (32 tests passed).
✨ **Result:** A cleaner codebase with centralized path management.

---
*PR created automatically by Jules for task [9523722981637211200](https://jules.google.com/task/9523722981637211200) started by @giauphan*

### DIFF
--- a/src/cli/commands.ts
+++ b/src/cli/commands.ts
@@ -10,6 +10,7 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as os from "os";
+import { getHermesConfigPath, getHermesPluginDir } from "../utils/pathUtils.js";
 import * as readline from "readline";
 
 const API_URL = process.env.CODEATLAS_API_URL || "https://your-server.com";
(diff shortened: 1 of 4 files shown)


giauphan:
cc @jules

This is a clean refactoring — I only see one minor thing worth flagging:

**Unused import risk**: The diff adds `getHomePath` to the `os` import line in `projectService.ts` but doesn't show `os` being removed. If `os.homedir()` was the only `os.*` call left in that file, the `* as os` import is now dead code and will produce a lint error. Same concern for `mcpServer.ts` if `os.homedir()` was the sole usage — though `os` may still be used elsewhere in those files (can't tell from the diff alone).

Everything else is fine:
- **Correctness**: `getHomePath()` is a direct pass-through to `os.homedir()`, so behavior is identical.
- **Security**: No injection vectors, no hardcoded secrets.
- **Performance**: Trivial — no impact.
- **Code quality**: Good — centralizing magic path strings into named functions eliminates duplication and improves discoverability.
- **Maintainability**: The thin `getHomePath()` wrapper enables future override via env var or config without touching every call site.

---

PR #65, opened 2026-07-15
⚡ Bolt: [Performance] Concurrent project scanning in enterprise vulnerabilities

giauphan:
💡 What: Changed the sequential `for...of` loop in `scan_enterprise_vulnerabilities` to a concurrent mapping using `Promise.all`.
🎯 Why: The loop sequentially awaited `loadAnalysisAsync` and `SecurityScanner.aiScan` for each project, causing linear time growth. Parallelizing these async calls significantly improves throughput.
📊 Measured Improvement: In a benchmark simulating 5 projects with dummy 100ms I/O delays, the execution time dropped from ~1004ms (sequential) to ~201ms (concurrent).

---
*PR created automatically by Jules for task [18155694100465037690](https://jules.google.com/task/18155694100465037690) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -4,3 +4,6 @@
 ## 2025-07-13 - [Performance improvement] Speed up code_search with early Regex check
 **Learning:** When performing full-text searches across many large files, calling `content.toLowerCase()` on the entire file contents to perform an `.includes()` check causes massive memory spikes due to allocating new strings. Using `new RegExp(escapedQuery, 'i')` allows for a case-insensitive existence check without duplicating the entire string memory, providing both speed and memory efficiency.
 **Action:** When filtering or performing early exit checks on large text blocks, prefer case-insensitive regex checks over converting the entire text to lowercase, to avoid excessive memory allocation and GC overhead.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

Nothing major to flag here. The change is clean and correct.

A few observations:

**Correctness**: The concurrent conversion is sound. Each iteration has its own try/catch, so one failure won't take down others. Null results from the `!loaded` early return are correctly filtered with `if (r)`. Error objects (always truthy) fall through correctly. No change in behavior.

**Security**: None. Same error messages as before.

**Performance**: This is a clear improvement for multi-project scans — `loadAnalysisAsync` runs in parallel rather than serially. The only caveat is that `projectsToScan` is already capped at `limit`, so unbounded concurrency isn't a concern here.

**Code quality**: The diff is minimal and focused. `scanPromises` is appropriately const and the post-filter loop is straightforward. No dead code or duplication introduced.

**One minor thing to consider**: If `loadAnalysisAsync` has internal caching or stateful side effects, concurrent calls could race on cache initialization. That's speculative without seeing its implementation, but worth verifying if you've seen flakiness in tests.

---

PR #66, opened 2026-07-15
🛡️ Sentinel: [HIGH] Fix authorization bypass in API

giauphan:
🚨 **Severity:** HIGH
💡 **Vulnerability:** The `checkAuth` function for multi-tenant authorization fell back to granting full 'enterprise' privileges via a mock local user if no authentication context was found.
🎯 **Impact:** When the software was deployed in multi-tenant mode (cloud), the mock security fallback intended for local development bypassed authentication completely, granting unauthorized requests full administrative capabilities.
🔧 **Fix:** Added a conditional check for the `CODEATLAS_MULTI_TENANT` deployment flag in `src/services/authService.ts`. If running in multi-tenant mode without valid auth context, it now throws an explicit `Unauthorized` error instead of returning local mock credentials.
✅ **Verification:** Verified by running `npm run build` and `npm run test` to ensure existing functionality remains working locally. Checked logic to ensure the `Unauthorized` error correctly gates access when `CODEATLAS_MULTI_TENANT` is enabled.

---
*PR created automatically by Jules for task [13244231966475233043](https://jules.google.com/task/13244231966475233043) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -10,3 +10,7 @@
 **Vulnerability:** The API key `CODEATLAS_API_KEY` was being redundantly sent as a query parameter (alongside the `x-api-key` header) in URLs within `src/services/dreamingService.ts` when making requests to CodeAtlas Cloud.
 **Learning:** Passing sensitive data like API keys in URL query parameters exposes them to server logs, browser history, and proxy servers (CWE-598).
 **Prevention:** Always transmit API keys and other sensitive credentials securely via HTTP headers (e.g., `x-api-key`, `Authorization: Bearer`), ensuring they are omitted from the request path.
+## 2026-07-15 - Authorization Bypass Risk in Local Mock Fallbacks
+**Vulnerability:** The `checkAuth` function used for multi-tenant authorization fell back to granting full 'enterprise' privileges via a mock local user if no authentication context was found.
+**Learning:** When software supports both single-tenant (local) and multi-tenant (cloud) deployment modes, mock security fallbacks designed for local development can inadvertently bypass authentication in production if they are not conditionally gated.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

This PR looks clean. The change is a focused security fix with no issues.

**Correctness:** Straightforward guard — if `contextAuth` is missing and multi-tenant mode is active, throw instead of silently granting enterprise access. The `=== "true"` strict comparison is good (avoids false positives from truthy env vars).

**Security:** Fixes exactly the vulnerability it describes. No injection vectors, no hardcoded secrets, no new attack surface.

**Performance:** Zero impact — a single string comparison on an early-return path.

**Code quality:** Minimal, readable, follows existing patterns. The error message is descriptive and actionable.

**Maintainability:** Self-documenting — the env var name and error message make the intent clear without needing a comment.

One minor thought: if `contextAuth` could be an empty object `{}` (falsy-ish but truthy in JS), this wouldn't catch it — but that's a pre-existing concern in the `contextAuth` source, not in this diff. As-is, the change is correct and sufficient.

---

giauphan:
cc @jules

This PR looks good overall — it's a targeted security hardening fix. A few observations:

**Correctness/Security**: The `Object.keys(contextAuth).length > 0` guard is a real improvement — an empty `{}` is truthy in JS and would previously bypass auth. Combined with the multi-tenant throw, this prevents the mock local-user fallback from silently granting enterprise privileges in production.

**Minor env-var robustness**: The check `=== "true"` is strict — someone setting `CODEATLAS_MULTI_TENANT=1` or just `CODEATLAS_MULTI_TENANT=` would silently get the local-user fallback. However, this is standard practice and well-documented in the sentinel, so it's acceptable.

**No issues with**: performance (negligible), code quality (minimal, focused change), maintainability (clear error message + documentation).

The PR is ready to merge.

---

giauphan:
✅ Merged in af1b8b2f

PR #68, opened 2026-07-16
⚡ Bolt: Optimize PhpParser string allocation and fs.readdir calls

giauphan:
💡 What: Replaced an inefficient `name !== name.toLowerCase().replace(/[^a-z]/g, '')` string allocation check in `PhpParser` with `/[^a-z]/.test(name)`. Upgraded `discoverProjects` and `discoverProjectsAsync` file system transversals in `projectService.ts` to use `fs.readdir` with `withFileTypes: true` to avoid excessive `fs.stat` calls.
🎯 Why: Creating a lowercase string and then running a regex replacement on it is inefficient, especially inside a regex capture loop, causing GC pressure. Checking file statuses individually using `fs.stat` is highly disk I/O heavy and slows down multi-project scanning. 
📊 Impact: Benchmark speedups show ~3x performance improvement inside PHP regex checks, and ~3x to ~22x improvement for asynchronous directory fetching. Reduces CPU and memory GC overhead when parsing PHP files with many function calls. Dramatically reduces disk I/O operations for enterprise codebases with many projects.
🔬 Measurement: Verified with local synthesized test harnesses using `tsx`. Confirmed that `test` checks continue to pass and output matches original behavior.

---
*PR created automatically by Jules for task [5539773824844697915](https://jules.google.com/task/5539773824844697915) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -12,3 +12,10 @@
 **Why:** Disk I/O bound nested loops are very inefficient, particularly when traversing potentially thousands of directories.
 **Impact:** Benchmark speedups show ~3x performance improvement. Specifically checking 2500 directories improved from 534ms to 173ms on local testing hardware.
 **Measurement:** Added custom `tsx` based scripts (`benchmark_scan.ts` and `benchmark_scan_opt.ts`) to synthesize an artificial tree with thousands of directories and compared total times before and after changes.
+## 2025-07-28 - [Performance improvement] Regex testing vs String manipulation
+**Learning:** Checking character matching (like finding non-lowercase alphabetical characters) by applying string allocations via `.toLowerCase().replace()` and comparing to the original is incredibly slow. Using a regular expression `.test()` without memory allocation provides almost 3x performance speedups inside parser loops.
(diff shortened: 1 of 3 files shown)


giauphan:
cc @jules

This PR looks clean. Here's my analysis:

**phpParser.ts** — The regex change is logically equivalent and correct. `/[^a-z]/.test(name)` matches exactly the same set as the old `name !== name.toLowerCase().replace(/[^a-z]/g, '')` idiom without allocating strings. Proper optimization.

**projectService.ts** — Both the sync and async `discoverProjects` changes are correct:
- `withFileTypes: true` eliminates the N stat calls per directory as the docs claim
- Symlink handling is right: `Dirent.isDirectory()` returns `false` for symlinks regardless of target, so the `isSymbolicLink()` fallback to `fs.statSync`/`fs.promises.stat` is necessary and correct
- Empty `catch` on stat failures for symlinks is fine (broken symlinks should be silently skipped)
- The `p.name` access matches the `Dirent` interface

No issues worth flagging. The only thing that stands out to me is that `p` is being used as a variable name for both `string` (before) and `Dirent` (after) — that's a bit confusing but harmless since the change is contained.

---

PR #69, opened 2026-07-17
⚡ Bolt: [Performance improvement] Optimize project discovery with fs.readdir withFileTypes

giauphan:
💡 What: The `discoverProjects` and `discoverProjectsAsync` functions used `fs.readdir` (which returned string names), and subsequently mapped those strings into absolute paths to check via N+1 synchronous (`fs.statSync`) and asynchronous (`fs.promises.stat`) `stat` operations. This optimization changes them to use `{ withFileTypes: true }`, bypassing the costly loop of disk calls.
🎯 Why: Iterating over many directories on an I/O constrained storage system blocks the node thread or overloads the I/O bus unnecessarily. Retrieving directory state alongside the read command scales exponentially better.
📊 Impact: This significantly cuts down the amount of disk I/O when discovering multi-tenant repositories in directories containing large numbers of users. Avoids N+1 system calls per project user space folder.
🔬 Measurement: Verify changes in execution length when running discovery across thousands of subdirectories by utilizing local timing functions. Ensure memory impact stays stable.

---
*PR created automatically by Jules for task [6302775531672019477](https://jules.google.com/task/6302775531672019477) started by @giauphan*

### DIFF
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -604,10 +604,18 @@ export function discoverProjects(tenantId?: string): { name: string; dir: string
       const userDir = path.join(tenantRoot, tenantId);
       if (fs.existsSync(userDir)) {
         try {
-          const userProjects = fs.readdirSync(userDir);
-          for (const p of userProjects) {
-            const fullPath = path.join(userDir, p);
-            if (fs.statSync(fullPath).isDirectory()) {
+          const userProjects = fs.readdirSync(userDir, { withFileTypes: true });
+          for (const entry of userProjects) {
+            const fullPath = path.join(userDir, entry.name);
+            let isDirectory = entry.isDirectory();
+            if (!isDirectory && entry.isSymbolicLink()) {
+              try {
+                isDirectory = fs.statSync(fullPath).isDirectory();
+              } catch {
+                continue;
+              }
+            }
+            if (isDirectory) {
               searchDirs.push(fullPath);
             }
           }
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

This looks clean. The change correctly handles symbolic links to directories by resolving them with `stat` when `isDirectory()` on the `Dirent` returns false for a symlink. Broken symlinks are safely skipped via `catch { continue }` / `catch { return }`.

No correctness, security, performance, or maintainability issues. The logic is consistent between the sync and async versions, and the error handling for broken symlinks is appropriate. Good change.

---

giauphan:
cc @jules

This looks clean. The key changes:

- **`withFileTypes: true`** avoids unnecessary `stat` calls for regular entries — a minor but real perf win.
- **Symlink resolution** is handled correctly: if the dirent says it's a directory, great; if it's a symlink, it stats the target and checks the resolved type.
- **Broken symlinks** are safely caught and skipped via the inner `catch { continue }` / `catch { return; }`.
- The sync and async variants are kept consistent.

No issues to flag.

---

giauphan:
cc @jules

This looks clean. The change properly handles symlinks in both the sync and async `discoverProjects` functions.

**Key observations:**

- **Correctness**: Solid. Using `withFileTypes: true` avoids an extra `stat` call per entry. For symlinks, it correctly falls back to `statSync`/`fs.promises.stat` (which follow symlinks) to determine if the target is a directory. Broken symlinks are caught and skipped via the empty `catch`.
- **Security**: No issues — `path.join` usage prevents path traversal, no new inputs introduced.
- **Performance**: Slightly improved — the old code called `stat` on *every* entry; now it only calls `stat` for symlinks, since `Dirent.isDirectory()` is instant.
- **Code quality**: The sync and async versions stay parallel in structure. The `return;` in the async catch produces `undefined` entries in the `Promise.all` array, but they're never used — harmless, and a common pattern.
- **Maintainability**: Error handling is appropriate — broken symlinks are silently skipped, preserving the existing behavior of ignoring inaccessible paths.

No bugs, no regressions. This is ready to merge.

---

giauphan:
cc @jules

This change looks good overall. Here's my analysis:

**Correctness** — The logic is sound. `readdir` with `withFileTypes: true` returns `Dirent` objects whose `isDirectory()` uses `lstat` semantics (doesn't follow symlinks). The code correctly handles that gap by checking `isSymbolicLink()` and falling back to `stat` (follows symlinks). Broken symlinks are caught and skipped. Nested symlinks work fine since `stat` follows the chain.

**Performance** — This is actually a *net improvement*. The old code called `stat` on every single entry. Now only symlinks need the extra syscall. Regular dirs and files are classified directly from the `readdir` dirent data.

**Security** — No concerns. Pure filesystem enumeration, no untrusted input being executed or exposed.

**Code quality** — Clean, consistent pattern between sync and async versions. The early `return` in the catch block is fine since `Promise.all` happily resolves `undefined` values. Minor nit: the async version returns `Promise<void>[]` where most elements are `undefined`, but since the results aren't used, it's harmless. Would be slightly cleaner to `.filter(Boolean)` or use `Promise.allSettled`, but not worth changing.

**Only real issue I'd flag:** The `catch { continue/return }` swallows all error types — not just `ENOENT` for broken symlinks but also `EACCES` (permission denied), `ELOOP` (too many symlink levels), `ENAMETOOLONG`, etc. If a user has a permissions issue on a directory, they'd silently get fewer projects listed with no indication why. A targeted `continue`/`return` on `ENOENT` specifically would be more honest. But this matches the existing error-handling style in the codebase, so I'd let it pass.

---

giauphan (commented):
@jules ⚠️ **Merge conflict** — `projectService.ts` needs rebase. Also `parser.ts.orig` and `parser.ts.rej` (patch artifacts from PR #75) should not be committed — add `*.orig` and `*.rej` to .gitignore.

giauphan:
Superseded — code already on main from PR #75 (same changes merged)

PR #70, opened 2026-07-17
🛡️ Sentinel: [CRITICAL] Fix YAML injection vulnerability in MCP setup

giauphan:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** YAML injection vulnerability in `src/presentation/mcpServer.ts` within the `setup_second_brain` tool. The `CODEATLAS_API_KEY` was being interpolated directly into a YAML string without escaping, which could allow a malicious user to inject arbitrary YAML nodes by providing an API key containing double quotes and newlines.
🎯 **Impact:** Structural configuration injection, which could lead to arbitrary command execution during MCP client configuration and startup.
🔧 **Fix:** Changed the interpolation of `key` from `"${key}"` to `${JSON.stringify(key)}`, which safely encodes the input string and neutralizes any structural characters.
✅ **Verification:** Verified that tests pass via `npm run test` and `npm run build` and recorded the finding in `.jules/sentinel.md`.

---
*PR created automatically by Jules for task [17839039152117133710](https://jules.google.com/task/17839039152117133710) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -14,3 +14,7 @@
 **Vulnerability:** The `checkAuth` function used for multi-tenant authorization fell back to granting full 'enterprise' privileges via a mock local user if no authentication context was found.
 **Learning:** When software supports both single-tenant (local) and multi-tenant (cloud) deployment modes, mock security fallbacks designed for local development can inadvertently bypass authentication in production if they are not conditionally gated.
 **Prevention:** Always check deployment/configuration flags (e.g., `CODEATLAS_MULTI_TENANT`) before returning mock credentials or bypassing security checks. Throw an explicit unauthorized error when running in a multi-tenant environment.
+## 2026-07-16 - YAML Injection in MCP Server Configuration
+**Vulnerability:** The `CODEATLAS_API_KEY` was being interpolated directly into a YAML string via a template literal (`"${key}"`), which exposed the configuration to YAML injection if the key contained double quotes and newline characters.
+**Learning:** Interpolating unvalidated input strings into YAML without proper escaping creates a risk of structural injection, potentially allowing an attacker to overwrite configuration blocks or execute arbitrary commands.
+**Prevention:** Always use `JSON.stringify(key)` instead of raw string interpolation (`"${key}"`) when injecting string values into generated YAML or JSON configuration blocks to securely escape quotes and newlines.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

The fix is correct and addresses the stated vulnerability well. A quick analysis:

**Correctness** — `JSON.stringify(key)` produces a valid JSON string literal (e.g. `"value with \"quotes\" and \nnewlines"`), and since YAML is a superset of JSON, this value is safely consumed as a YAML scalar without structural injection. The outer double quotes from the original template are no longer needed since `JSON.stringify` adds them — and the diff correctly removes them. ✓

**Security** — The raw interpolation `"${key}"` could let a key value like `foo"\n  injected: true\n` mutate the YAML structure. `JSON.stringify` escapes both `"` → `\"` and embedded newlines → `\n` sequences, preventing injection. ✓

**Performance / Code quality / Maintainability** — No issues. The inline comment explains *why* `JSON.stringify` is used, which is exactly the right level of documentation for a security-sensitive escape decision. No dead code, no duplication, no anti-patterns.

Everything looks good — the PR is a clean, minimal fix that eliminates a concrete injection vector with the right escaping mechanism.

---

giauphan:
cc @jules

This looks clean. The fix replaces raw template interpolation (`"${key}"`) with `JSON.stringify(key)`, which properly escapes embedded double quotes, newlines, and other special characters that could break the YAML structure. This directly addresses the documented injection vector.

One minor edge case: if `key` were ever `undefined`, `JSON.stringify` returns the primitive `undefined` (not the string), which would produce `CODEATLAS_API_KEY: undefined` in the YAML — a silent misconfiguration. But for an API key parameter that should always be a string, this is fine in practice.

---

giauphan:
cc @jules

The fix looks correct and complete. `JSON.stringify(key)` handles all the edge cases — double quotes get escaped as `\"`, newlines become `\n`, and the output is already wrapped in double quotes, producing valid YAML regardless of what characters the key contains. 

Good call on documenting the pattern in sentinel.md too. No other issues here.

---

giauphan (commented):
@jules ✅ YAML injection fix in `mcpEntry` looks correct (`JSON.stringify(key)`). 

However — I also checked the Python plugin template (line 2026-`KEY = os.environ.get(...)`) — this reads from env at runtime, NOT string interpolation, so the CRITICAL injection concern is NOT applicable. Only the YAML template was vulnerable. Fix confirmed complete.

giauphan:
Superseded by #78 (already merged — same YAML injection fix)

PR #71, opened 2026-07-18
⚡ [performance] Optimize config file checking to prevent event loop blocking

giauphan:
💡 **What:** Replaced synchronous `fs.existsSync` inside a for-loop with `Promise.all` and `fs.promises.access` to check for config files asynchronously and concurrently.
🎯 **Why:** To prevent blocking the Node.js event loop. The original code executed synchronous I/O operations in a loop, which delays the processing of other concurrent requests and tasks in a server environment. This optimization ensures that checking for project configuration files doesn't stall the event loop.
📊 **Measured Improvement:** 
Due to OS caching and the nature of the development environment's fast local disk, synthetic benchmarks (tight loops checking for non-existent files) showed a lower raw execution time for the synchronous version simply because it didn't incur the overhead of Promises and event loop scheduling.
*   **Sync (Baseline in tight loop):** Duration: 158 ms, Max Event Loop Lag: 0.00 ms
*   **Async (in tight loop):** Duration: 2865 ms, Max Event Loop Lag: 3.21 ms

*However*, this raw throughput in a tight synthetic loop is misleading for a web server scenario. The rationale for this change is to improve **concurrency**. While a single request's I/O operation might be microscopically faster with sync I/O on a cached file, sync I/O fundamentally blocks the entire Node.js process. In a real-world scenario (slower disks, network drives, concurrent user requests), the async approach is vital because it allows the MCP server to handle other tasks while waiting for file system responses, which is a standard best practice in Node.js performance engineering. The optimization trades a tiny amount of per-request overhead for significantly better overall server responsiveness under load.

---
*PR created automatically by Jules for task [12213998591673407310](https://jules.google.com/task/12213998591673407310) started by @giauphan*

### DIFF
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1822,9 +1822,16 @@ export function registerTools(server: McpServer) {
 
       // Config files
       ctx.configFiles = {};
-      for (const [key, f] of Object.entries({ tsconfig: "tsconfig.json", eslint: ".eslintrc.js", prettier: ".prettierrc", jest: "jest.config.js", vitest: "vitest.config.ts", playwright: "playwright.config.ts", docker: "Dockerfile" })) {
-        ctx.configFiles[key] = fs.existsSync(path.join(projectDir, f));
-      }
+      await Promise.all(
+        Object.entries({ tsconfig: "tsconfig.json", eslint: ".eslintrc.js", prettier: ".prettierrc", jest: "jest.config.js", vitest: "vitest.config.ts", playwright: "playwright.config.ts", docker: "Dockerfile" }).map(async ([key, f]) => {
+          try {
+            await fs.promises.access(path.join(projectDir, f));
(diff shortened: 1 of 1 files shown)


giauphan:
cc @jules

This is a clean, straightforward improvement. Everything looks good:

- **Correctness**: ✓ The try/catch correctly maps `access` resolution/rejection to `true`/`false`, matching the original `existsSync` behavior. No race conditions since each iteration writes to a different key.
- **Performance**: ✓ Converts 7 sequential blocking `existsSync` calls into parallel async `access` calls. Strictly better.
- **Security / Code quality / Maintainability**: No issues. Idiomatic async pattern.

The only subtle difference worth noting: `fs.promises.access` checks **accessibility** (file exists + readable permissions), while `existsSync` checks only **existence**. In practice these config files are always readable if they exist, so this shouldn't matter — and arguably `access` is more correct since we care about the file being usable.

---

PR #83, opened 2026-07-21
⚡ Bolt: [performance improvement] Optimize getDirContents to eliminate fs.statSync loops

giauphan:
💡 What: Refactored `getDirContents` in `src/analyzer/parser.ts` to use `fs.readdirSync` with `withFileTypes: true` and replaced the `Set<string>` directory cache with a `Map<string, boolean>` to store the `isDirectory` status of files.
🎯 Why: Subsequent synchronous `fs.statSync()` calls when traversing code trees create significant disk I/O bottlenecks. Caching `isDirectory` statuses directly from directory listing avoids these bottlenecks.
📊 Impact: Expected to greatly reduce the number of synchronous disk I/O calls when resolving thousands of ES module imports during project analysis.
🔬 Measurement: Run project analysis on a large repository and measure the reduction in CPU and disk time spent on filesystem API calls.

---
*PR created automatically by Jules for task [7218446011652933986](https://jules.google.com/task/7218446011652933986) started by @giauphan*


giauphan:
cc @jules

Looks like `escapeRegExp` is used but not defined or imported in `mcpServer.ts`. Let me verify.Here's my review of this PR.

**Critical bug: `escapeRegExp` is undefined in `mcpServer.ts`**

The new code calls `escapeRegExp(base)` but this function is never defined or imported anywhere in the codebase. This will throw a `ReferenceError` at runtime. It needs to be implemented (escape `.\^$*+?{}[]|()` chars) or the simpler `.toLowerCase().includes()` approach should stay.

**Parser.ts changes look good**

- Changing `Set<string>` to `Map<string, boolean>` and using `withFileTypes: true` correctly eliminates the redundant `fs.statSync` calls as intended.
- The symlink fallback to `fs.statSync` is handled properly — if stat fails, `isDirectory` stays `false` and the entry is still cached.
- On the caller side, `parentContents.get(bName) === true` correctly handles missing entries (`undefined !== true`) and non-directory files (`false !== true`). No issues.

**Bolt.md is just documentation** — fine.

**Summary**: One real bug (undefined `escapeRegExp`), rest is clean.

---

giauphan:
cc @jules

I've reviewed the diff. This is a clean, well-scoped performance optimization. No bugs, security issues, or quality problems to flag.

The key change: `getDirContents` now uses `fs.readdirSync(dirPath, { withFileTypes: true })` to capture both filenames and their directory/symlink status in a single I/O pass, stored as `Map<string, boolean>`. The downstream use site trades a `fs.statSync` call for a `Map.get()` lookup.

A couple minor observations that don't rise to actionable issues:

- The old `try` block wrapping the replacement code is now vestigial — a `Map.get()` can't throw. Harmless, no need to restructure.
- Symlink resolution (`isSymbolicLink()` → fallback `fs.statSync`) happens eagerly for *all* symlinks during the directory scan, whereas before it was lazy (only on access). In practice this is a wash or still a net win, since you'd stat them anyway.

Everything looks good, no issues worth raising.

---

giauphan:
Superseded by PR #84 (which is already merged).

PR #85, opened 2026-07-22
🔒 Prevent path traversal in project loading and command execution

giauphan:
🎯 **What:** 
Fixed a critical security vulnerability involving path traversal and potential command execution outside user's authorized workspaces in the MCP tool handlers.

⚠️ **Risk:** 
If left unfixed, an attacker could supply an arbitrary directory path as `project` to `loadAnalysisAsync` or `run_script`, and the server would resolve the path and potentially execute commands (e.g. `npm build`, `git status`) in unauthorized directories (such as system paths or other tenant's environments). This can result in unauthorized reading of files, sensitive data exposure, and arbitrary command execution.

🛡️ **Solution:** 
- Enforced boundary checks inside `loadAnalysisAsync` (in `src/services/projectService.ts`) to ensure that any target directory falls within the list of authorized projects returned by `discoverProjectsAsync`.
- Replicated this robust security boundary check directly into the `run_script` tool in `src/presentation/mcpServer.ts` for defense in depth.

---
*PR created automatically by Jules for task [13559946412106956244](https://jules.google.com/task/13559946412106956244) started by @giauphan*

### DIFF
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -7,6 +7,7 @@ import { getHomePath, getHermesConfigPath, getHermesPluginDir, getClaudeConfigPa
 import { checkAuth, logActivity } from "../services/authService.js";
 import {
   discoverProjectsAsync,
+  isPathInAuthorizedProjects,
   loadAnalysisAsync,
   getStats,
   fileExists,
@@ -1930,12 +1931,7 @@ export function registerTools(server: McpServer) {
 
       // Ensure the project directory is an authorized workspace to prevent path traversal
       const authorizedProjects = await discoverProjectsAsync(auth.uid);
-      const isAuthorized = authorizedProjects.some(p => {
-        const resolvedProj = fs.realpathSync(path.resolve(projectDir));
-        const resolvedAuth = fs.realpathSync(path.resolve(p.dir));
-        return resolvedProj === resolvedAuth || resolvedProj.startsWith(resolvedAuth + path.sep);
-      });
-      if (!isAuthorized) {
+      if (!isPathInAuthorizedProjects(projectDir, authorizedProjects)) {
         return { content: [{ type: "text" as const, text: JSON.stringify({ error: "Unauthorized project directory" }) }] };
       }
 
@@ -2021,12 +2017,7 @@ export function registerTools(server: McpServer) {
       // 🛡️ Sentinel Security Validation
       // Ensure the project directory is an authorized workspace to prevent path traversal
       const authorizedProjects = await discoverProjectsAsync(auth.uid);
-      const isAuthorized = authorizedProjects.some(p => {
-        const resolvedProj = fs.realpathSync(path.resolve(projectDir));
-        const resolvedAuth = fs.realpathSync(path.resolve(p.dir));
-        return resolvedProj === resolvedAuth || resolvedProj.startsWith(resolvedAuth + path.sep);
-      });
-      if (!isAuthorized) {
+      if (!isPathInAuthorizedProjects(projectDir, authorizedProjects)) {
         return { content: [{ type: "text" as const, text: JSON.stringify({ error: "Unauthorized project directory" }) }] };
       }
 
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1930,9 +1930,11 @@ export function registerTools(server: McpServer) {
 
       // Ensure the project directory is an authorized workspace to prevent path traversal
       const authorizedProjects = await discoverProjectsAsync(auth.uid);
-      const isAuthorized = authorizedProjects.some(p =>
-        projectDir === p.dir || projectDir.startsWith(p.dir + path.sep)
-      );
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

I found a real issue with the path traversal guard and a couple of observations worth noting.

**1. Path traversal bypass — `startsWith` without normalization**

Both checks assume `projectDir`/`absPath` and `p.dir` are clean, canonical paths. But if the input path contains `..` segments (e.g. `/authorized-project/../../etc/passwd`), `startsWith` passes while the resolved path exits the workspace. Example:

```
projectDir = "/home/user/project/../../etc"
projectDir.startsWith("/home/user/project/")  // true — bypass!
```

Both locations should normalize paths with `path.resolve()` (and ideally `fs.realpathSync()` for symlink hardening) before the comparison. A shared helper would also prevent this from drifting apart.

**2. Missing `.some()` short-circuit on `p.dir === projectDir`**

In `mcpServer.ts`, the exact-match check `projectDir === p.dir` is redundant with the `startsWith(p.dir + path.sep)` check — if they're equal, `startsWith` already catches it (well, it won't because there's no separator...). Actually, the `===` handles the exact-match case correctly since `projectDir.startsWith(p.dir + path.sep)` is false when they're equal (no trailing `/`). So the two conditions correctly cover both the exact match and the subdirectory case — this is fine as written.

**3. Performance concern — `discoverProjectsAsync` on every call**

If `discoverProjectsAsync(auth.uid)` queries a DB or scans the filesystem, calling it on every security validation request adds latency. Consider caching the authorized project list (with TTL) or passing it as part of the request context.

**4. Symlink attack surface**

Even with `startsWith` fixed using `path.resolve`, `fs.realpathSync` is needed to resolve symlinks. If an authorized project dir contains a symlink like `proj/link -> ../../unauthorized`, paths through that symlink would pass the canonical-path check but lead outside the intended boundary. Use `fs.realpathSync` (or the async equivalent) on both sides.

**Summary**: The logic is directionally correct, but the path traversal guard is incomplete without `path.resolve()`. I'd consolidate into a helper:

```typescript
function isAuthorizedPath(targetPath: string, authorizedDirs: string[]): boolean {
  const resolved = fs.realpathSync(path.resolve(targetPath));
  return authorizedDirs.some(dir => {
    const canonicalDir = fs.realpathSync(path.resolve(dir));
    return resolved === canonicalDir || resolved.startsWith(canonicalDir + path.sep);
  });
}
```

This handles `..`, symlinks, and the exact-match + subdirectory cases in one pass.

---

giauphan:
cc @jules

I see three related changes that add path traversal prevention via `discoverProjectsAsync` + `fs.realpathSync`. Here are the issues:

1. **Missing error handling around `fs.realpathSync` in mcpServer.ts** — both new blocks call `fs.realpathSync` without a try/catch. If `projectDir` or `p.dir` doesn't exist (symlink dead, deleted between discovery and validation), it throws ENOENT and crashes the handler. The `projectService.ts` change does this correctly with a try/catch — the mcpServer.ts changes should match.

2. **Code duplication** — the authorization check pattern (`realpathSync` + `some` + `startsWith`) appears 3 times now (twice in mcpServer.ts, once in projectService.ts). This should be extracted into a shared helper like `isPathInAuthorizedProjects(resolvedPath, projects): boolean`. If the edge-case fix from #1 needs to change, there are 3 places to update.

3. **Minor**: In `projectService.ts`, the `console.error` logging of the rejected `absPath` could leak directory structure info in environments where that's sensitive — but this is a debug log, so probably fine.

Everything else looks reasonable — using `realpathSync` over string comparison is the right fix for symlink-based traversal, and the overall approach is sound.

---

giauphan:
cc @jules

Good PR. Clean extraction of a path-traversal guard into a reusable function with consistent usage. Here's my review:

**Correctness** – Looks solid. `realpathSync` resolves symlinks and dot-path components, so path-traversal via `../` or symlink chains is prevented. The `startsWith(auth + sep)` check correctly catches children without false-matching prefix collisions (e.g. `/home/a` won't match `/home/abc` because of the separator). The order in `loadAnalysisAsync` — check authorization *before* `registerProject` — is right.

**Security** – No issues. The function is conservative by design: if `realpathSync` throws (missing path, permissions error), it returns `false` (deny). This avoids accidentally authorizing a path just because resolution failed.

**Performance** – Fine. Sync `realpathSync` calls on a handful of paths in tool handlers and the auto-scan path — no perf concern.

**Code quality** – Good DRY refactor. The old inline logic in `mcpServer.ts` is replaced by a semantically identical call to `isPathInAuthorizedProjects`. The new guard in `loadAnalysisAsync` adds a useful defense-in-depth layer that was missing before.

**One minor thing** – The function uses sync filesystem calls (`fs.realpathSync`) while the rest of `projectService.ts` exposes `fsWrapper` (presumably for testability). If testing `isPathInAuthorizedProjects` directly matters, the `fs.realpathSync` call may need to go through a wrapper. As-is, it's consistent with the rest of the sync calls in this file — not a blocker.

Everything else looks good, no blockers.

---

PR #86, opened 2026-07-22
🔒 Fix indirect command injection in run_script tool via args

giauphan:
🎯 **What:** Fixed an indirect command injection vulnerability in the `run_script` tool by blocking shell metacharacters in the `args` parameter.

⚠️ **Risk:** Although `spawnSync` was used with `shell: false`, arguments passed to `npm run` are evaluated by the shell when npm executes the actual target script. This allowed an attacker to inject shell commands via the args string. If exploited, it could lead to arbitrary command execution on the host machine.

🛡️ **Solution:** Implemented a regex allowlist (`/[&|;<>$`\n\r]/`) that inspects the `args` parameter. If any forbidden shell metacharacter is detected, it returns an explicit error message instead of executing the command, effectively neutralizing the attack vector while preserving normal argument functionality.

---
*PR created automatically by Jules for task [16420529453983174616](https://jules.google.com/task/16420529453983174616) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -18,3 +18,7 @@
 **Vulnerability:** The `CODEATLAS_API_KEY` was directly interpolated into a YAML string using string interpolation and double quotes (`"${key}"`).
 **Learning:** Directly embedding strings into structured data formats like YAML or JSON can lead to injection vulnerabilities if the string contains quotes or newlines.
 **Prevention:** Always use `JSON.stringify(value)` to securely escape variables when generating YAML or JSON configuration strings programmatically.
+## 2026-07-22 - [Indirect Command Injection]
+**Vulnerability:** Indirect command injection vulnerability in `run_script` tool via arguments passed to `npm run` which executes the target script in a shell environment.
+**Learning:** Even when `spawnSync` is used with `shell: false`, passing untrusted arguments to commands like `npm run` (which in turn launch subshells) can lead to command injection if the arguments are appended directly to the target script and contain shell metacharacters.
+**Prevention:** Strictly validate or sanitize untrusted input destined for command-line arguments by blocking shell metacharacters (`&|;<>$`\n\r`) before passing them to process executors.
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -23,6 +23,8 @@ function escapeRegExp(str: string): string {
   return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+const SHELL_METACHAR_RE = /[&|;<>$`\\\n\r]/;
+
 export function registerTools(server: McpServer) {
   // Tool -1: Analyze a project
   server.tool(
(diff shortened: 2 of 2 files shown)


giauphan:
cc @jules

This looks good to me. A clean, focused security fix.

The regex correctly blocks the shell metacharacters (`&`, `|`, `;`, `<`, `>`, `$`, backtick, `\`, `\n`, `\r`) that could be used for indirect command injection through `npm run`'s subshell. The sentinel entry accurately documents the vulnerability mechanism (spawnSync with `shell: false` bypassed by npm's own shell invocation) and the prevention guidance is correct.

Only minor observation: `*`, `?`, `(`, `)`, `'`, `"` aren't blocked but they're lower-risk for command injection in this context — globbing and subshells are limited without `$`, and quotes primarily break argument boundaries rather than injecting new commands. The set chosen is the right tradeoff for this use case.

Everything looks correct and mergable.

---

giauphan:
cc @jules

The change looks solid overall. A few observations:

**Correctness:** The regex is unanchored (no `^`/`$`), which is actually **correct here** — you want to reject if *any* metacharacter appears anywhere in the string, not just reject strings that are *entirely composed of* metacharacters. No false-negative risk for the targeted characters.

**Security:** Covers the main injection vectors (`&`, `|`, `;`, `<>`, `$`, `` ` ``, `\`, `\n`, `\r`) that an `npm run` subshell could interpret. One gap: `(` and `)` aren't blocked, so `$(cmd)` (already caught by `$`) is fine, but a bare subshell like `$(cmd)` is caught. No real gap here.

**Performance:** Fine — single regex test per invocation. Minor nit: the regex is re-created on every call. Extract it to a module-level `const`:

```ts
const SHELL_METACHAR_RE = /[&|;<>$`\\\n\r]/;
```

**Code quality / UX:** The error response doesn't tell the caller *which* characters are forbidden or what they passed. A more actionable message would be helpful:
```
"Security Error: Arguments contain forbidden shell metacharacters (& | ; < > $ ` \\). Received: <truncated input>"
```

But that's a polish point, not a correctness issue. The fix addresses the documented vulnerability correctly.

---

giauphan:
cc @jules

I see one significant issue with this change — the metacharacter block is **overly broad** and will cause false positives for legitimate arguments.

**The problem:** Characters like `$`, `>`, `<`, `|`, and `;` are common in perfectly valid npm script arguments:
- `$` — template strings, env variable names passed to scripts
- `>` / `<` — comparison operators in test descriptions (e.g., `"should return x > y"`)
- `|` — argument separators in many CLI tools
- `;` — command separators, or used in test/sample data

For example, these legitimate invocations would be blocked:
```
"name=foo|bar"      (pipe in a value)
"description: a > b"  (comparison in a test title)
"--template $USER"    (env var reference)
```

The actual attack surface is narrower: `npm run` does invoke a subshell via its script runner, but the arguments are passed as positional `argv` to the script being run. The injection risk depends on whether the *script itself* unsafely interpolates those args into a shell command. Blocking on the MCP server side is a reasonable defense-in-depth move, but this regex is too aggressive.

**My recommendation:** Either:
1. Escape these characters instead of rejecting (best for UX)
2. Only block the characters that are truly dangerous in this specific context (backtick, `$()` for command substitution, `;` for command chaining at the script level), and let safe ones like `>`, `<`, `|` through

Other than that, the code is clean — the error format matches existing patterns, the regex is compiled once, and the truncation in the error message is a nice touch.

---

PR #95, opened 2026-07-24
🛡️ Sentinel: [High] Path Traversal in git_changes

giauphan:
🎯 **What:** Fixed a path traversal security vulnerability in the `git_changes` tool that allowed unauthorized execution of git commands.
⚠️ **Risk:** The previous implementation failed to resolve paths with `../` prior to checking authorization boundaries. Attackers could theoretically manipulate `loaded.projectDir` to perform path traversal and read git histories from arbitrary directories outside of authorized workspaces (e.g. system files or other tenant workspaces).
🛡️ **Solution:** The path is now strictly resolved using `fs.realpathSync()` *before* being validated by `isPathInAuthorizedProjects()`. This ensures the validation evaluates the final canonical path. The resolved path is then explicitly used for subsequent `child_process.execFileSync` git commands to prevent traversal via `cwd` context.

---
*PR created automatically by Jules for task [13505525526624169174](https://jules.google.com/task/13505525526624169174) started by @giauphan*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -26,3 +26,7 @@
 **Vulnerability:** MCP tools allowed unbounded string lengths in Zod schemas (`z.string()`).
 **Learning:** Missing input length validation allows attackers to submit extremely large strings causing excessive memory usage and parsing overhead (CWE-400), leading to Denial of Service (DoS).
 **Prevention:** Always explicitly set maximum string lengths (e.g., `.max(255)`) on all `z.string()` schemas.
+## 2026-07-28 - [Path Traversal in Workspace Validation]
+**Vulnerability:** Path traversal in `git_changes` allowed attackers to execute git commands on unauthorized directories by bypassing the `isPathInAuthorizedProjects` check using unresolved paths containing `../`.
+**Learning:** Resolving a user-provided path via `fs.realpathSync()` *after* a security authorization check has already executed is ineffective. The boundary verification (`startsWith`) must evaluate the final, fully-resolved canonical path.
+**Prevention:** Always resolve file system paths fully using `fs.realpathSync()` **before** performing any authorization or boundary validation checks.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

This is a clean, well-structured security fix. No blocking issues.

**Key observations:**

- **Correctness** — The fix is correct: `fs.realpathSync()` is called **before** `isPathInAuthorizedProjects()`, which is the right ordering. All subsequent git commands consistently use `resolvedDir`. The sentinel entry's diagnosis ("resolve before authorize") is accurate.

- **Security** — Properly closes the path traversal gap for the `git_changes` tool. One edge worth noting: `realpathSync` resolves symlinks, so a symlink inside an authorized directory pointing outside would now be accepted — but that's standard POSIX behavior and generally preferred over blocking valid symlink setups.

- **Performance** — No concerns. Single `realpathSync` call, git commands unchanged.

- **Code quality** — Minimal, focused change. Dead code (`projectDir`) properly removed, all references migrated to `resolvedDir`.

- **Maintainability** — The sentinel entry documents why the fix exists, which helps prevent regression. Error handling for invalid project directories returns a consistent JSON error response.

One minor observation: `resolvedDir` is initialized as `""` and the code assumes the try block either sets it or returns early. This is safe here (the catch returns), but it's a fragile pattern — a future refactor that removes the early return would silently propagate an empty string. Initializing as `undefined` and adding a type guard would be more robust, though not worth blocking on.

---

giauphan:
cc @jules

The fix is correct — `resolvedDir` is now derived from `fs.realpathSync()` **before** the `isPathInAuthorizedProjects` call, which closes the path traversal gap. All subsequent `cwd` usages and the `existsSync(.git)` check also use the resolved path. Error handling for the realpath failure is present and consistent with the existing error-response pattern.

No issues to flag here. The sentinel document entry (dated 2026-07-28) references a future date relative to the session context (2026-07-24), but that's a documentation date, not a code concern — everything else looks good.

---

PR #113, opened 2026-07-30
⚡ Bolt: [Performance improvement] Optimize Jaccard similarity computation

giauphan:
💡 **What:** Refactored the Jaccard similarity computation in `src/presentation/mcpServer.ts` (`detect_code_similarities`) to calculate `intersectionSize` and `unionSize` manually via a loop and counter, instead of using `[...a.tokens].filter(...)` and `new Set()`.
🎯 **Why:** The existing approach was creating large arrays and sets for every single pair checked inside an $O(N^2)$ loop (up to 44,850 pairs). This caused massive memory allocations and Garbage Collection (GC) pressure, blocking the main thread and slowing down the tool.
📊 **Impact:** Reduces memory allocations in this loop from $O(\text{Pairs} \times \text{Tokens})$ down to $O(1)$ extra space, preventing GC pauses and significantly speeding up the `detect_code_similarities` tool execution.
🔬 **Measurement:** Verify by running the `detect_code_similarities` MCP tool on a large project and observing reduced memory usage and faster execution times compared to the previous version. Tests pass successfully.

---
*PR created automatically by Jules for task [18075586700359120433](https://jules.google.com/task/18075586700359120433) started by @giauphan*


giauphan (commented):
cc @jules

This looks correct and clean. Let me verify the math:

- **Intersection**: `intersectionSize` counts tokens in `a` that also exist in `b` — same result as the old `intersection.size`.
- **Union**: `|A ∪ B| = |A| + |B| - |A ∩ B|` via inclusion-exclusion — mathematically identical to the old `union.size` without allocating a combined Set.
- **Edge case**: Both versions skip when `unionSize === 0` (both sets empty), avoiding division by zero.
- **Output fields**: `sharedTokens` and `totalTokens` consistently use the counter variables instead of `.size`.

No bugs, no regressions, no security concerns. This is a straightforward performance optimization that eliminates temporary Array and Set allocations in an O(N²) similarity loop. The intermediate `intersectionSize` and `unionSize` are plain numbers, which means zero GC pressure compared to the original which allocated a new Array and up to two new Sets per iteration.

Only minor nit: if `thresh` is very low and the loop body executes for many pairs, this change yields proportionally more benefit — could be worth calling that out in the bolt.md entry, but what's there already covers the key insight.

giauphan:
cc @jules

This looks correct and clean. Let me verify the math:

- **Intersection**: `intersectionSize` counts tokens in `a` that also exist in `b` — same result as the old `intersection.size`.
- **Union**: `|A ∪ B| = |A| + |B| - |A ∩ B|` via inclusion-exclusion — mathematically identical to the old `union.size` without allocating a combined Set.
- **Edge case**: Both versions skip when `unionSize === 0` (both sets empty), avoiding division by zero.
- **Output fields**: `sharedTokens` and `totalTokens` consistently use the counter variables instead of `.size`.

No bugs, no regressions, no security concerns. This is a straightforward performance optimization that eliminates temporary Array and Set allocations in an O(N²) similarity loop. The intermediate `intersectionSize` and `unionSize` are plain numbers, which means zero GC pressure compared to the original which allocated a new Array and up to two new Sets per iteration.

Only minor nit: if `thresh` is very low and the loop body executes for many pairs, this change yields proportionally more benefit — could be worth calling that out in the bolt.md entry, but what's there already covers the key insight.

---

giauphan (commented):
cc @jules

⚠️ Auto-review skipped: command-code credits exhausted. Top up at https://commandcode.ai to enable reviews.

giauphan:
cc @jules

⚠️ Auto-review skipped: command-code credits exhausted. Top up at https://commandcode.ai to enable reviews.

---

giauphan (commented):
cc @jules

⚠️ Auto-review skipped: command-code credits exhausted. Top up at https://commandcode.ai to enable reviews.

giauphan:
cc @jules

⚠️ Auto-review skipped: command-code credits exhausted. Top up at https://commandcode.ai to enable reviews.

---

PR #133, opened 2026-08-07
🔒 Sentinel: High - Exposure of API Key in MCP Server Configuration

giauphan:
🎯 **What:** The `setup_second_brain` tool previously wrote the `CODEATLAS_API_KEY` in plaintext directly into the generated MCP client configuration files (for Claude Desktop and Hermes).

⚠️ **Risk:** Hardcoding secrets inside application configuration files exposes the credentials to potential leakage. If these configuration files were shared, backed up insecurely, or accessed by a malicious script running on the machine, the CodeAtlas cloud account could be compromised.

🛡️ **Solution:** The tool now stores the API key securely in an application-specific environment file (`~/.codeatlas/.env`) with restricted file permissions (`0o600`), and removes the hardcoded keys from the generated MCP client configurations. The main server entrypoint (`index.ts`) has been updated to automatically load this global configuration file on startup, ensuring the server can still authenticate seamlessly without exposing the key in plaintext configs.

---
*PR created automatically by Jules for task [686643087161341349](https://jules.google.com/task/686643087161341349) started by @giauphan*


giauphan:
## 🤖 Automated PR Review (giauphan/codeatlas-mcp-server #133)

Looking at the changes to understand what's being done and flag any issues.

**Summary:** This PR moves `CODEATLAS_API_KEY` out of plaintext MCP client configs into a secure `~/.codeatlas/.env` file, with `index.ts` loading it dynamically via dotenv. Good security direction overall.

**Issues found:**

1. **Race condition in `.env` write logic** (`mcpServer.ts:2238-2250`) — The read-then-write pattern is not atomic. Two concurrent invocations could both read stale content, both append/replace, and one write silently overwrites the other. Not catastrophic here since the key value is the same, but still a data race on the file.

2. **`fs.readFileSync` without error handling** (`mcpServer.ts:2240`) — Inside the `try/catch`, reading an existing `.env` file is fine if the file exists, but if something changes between `existsSync` check and `readFileSync` (TOCTOU), the inner catch won't be reached because we're already inside the outer `catch`. Minor edge case though.

3. **Regex replacement can corrupt lines** (`mcpServer.ts:2245`) — The regex `/CODEATLAS_API_KEY=.*(\r?\n|$)/g` greedily matches everything after `=` including comments or trailing text on that line. If someone has `CODEATLAS_API_KEY=old_value # note`, the `# note` part is preserved in `$1` which is correct, but if the line doesn't end with newline (last line in file), `$1` is empty and no newline is appended — meaning subsequent writes may never terminate that line properly. Should ensure a trailing newline regardless.

4. **File permissions on parent directory missing** (`mcpServer.ts:2236`) — Creates `~/.codeatlas` with `mode: 0o700`, good. But doesn't verify the path is actually within the home directory. An attacker who can influence `os.homedir()` return value could redirect writes elsewhere. Low risk in practice but worth noting.

5. **No cleanup of old config references** — The change removes `env` from the MCP config entries written to Claude Desktop and Hermes configs, which is correct. But if a previous version already wrote those files with env vars embedded, old entries persist until the next time the tool runs and overwrites those sections. Consider adding explicit removal of any leftover `env` blocks in previously-written configs.

**Overall:** Solid improvement to credential handling. The main actionable fix is ensuring the `.env` write appends a newline even when replacing the last line without a trailing newline (issue #3).

---
*Reviewed by github-review-bot*

giauphan (commented):
cc @jules

**What this PR does:** Migrates `CODEATLAS_API_KEY` from plaintext embedding in MCP config files (Hermes, Claude Desktop) to a secure `~/.codeatlas/.env` file with `0o600` permissions, then loads it globally via `dotenv`. Also updates sentinel.md and cleans up legacy embedded keys during migration.

### Issues

1. **Race condition on env file writes** — Two concurrent invocations of `setup_second_brain` could both read the existing `.env`, one overwrites, the other writes back its stale+modified content, silently dropping the first write's key. The comment even says "partial mitigate." If you want to be stricter, use a lock file or atomic temp-file + rename. Low urgency given this is a local dev tool.

2. **Hermes cleanup regex is fragile** — `[ \t]*CODEATLAS_API_KEY:.*(\r?\n|$)` will strip any line containing that pattern, which is fine, but the follow-up empty-env-block cleanup (`[ \t]*env:[ \t]*(\r?\n)(?![ \t]+[A-Za-z0-9_]+:)`) has an odd negative lookahead that only prevents stripping if there's *another* env entry on the next line. If `env:` was the last key in the block, it gets removed. This seems correct but the regex is non-obvious — a comment explaining intent would help. More importantly, the cleanup runs before checking whether `codeatlas:` exists at all, so it can strip unrelated `env:` blocks if the user happened to have a similar pattern nearby. Should guard cleanup behind detecting the `codeatlas:` section specifically.

3. **Claude Desktop `delete` doesn't preserve structure safely** — After deleting empty `env` objects, if either server's only remaining property was `env`, the entire server entry becomes `{ command: "...", args: ["..."] }` which is fine. But if some other property existed (unlikely), `Object.keys(...).length === 0` correctly cleans up. This is OK.

4. **Missing error handling for home directory** — `os.homedir()` can return `null` on some systems. `path.join(null, ".codeatlas")` produces `.codeatlas` (current dir), which writes without permission checks. The `getHomePath()` call already guards this, but `index.ts` uses `os.homedir()` directly without the same check. Should wrap in a try/catch or use `getHomePath()` consistently.

5. **Sentinel.md reordering** — The entries were reordered by date. Not a bug, but worth noting that chronologically sorting them would make the file easier to scan. Currently it goes 2025-02 → 2024-08 → 2025-02.

### Green flags

- File permissions (`0o700` for dir, `0o600` for env) are correct.
- The `override: false` in `dotenv.config()` is right — don't clobber already-set env vars.
- Cleaning up legacy `env: { CODEATLAS_API_KEY: ... }` references in existing configs prevents dangling secrets.
- Sentinel.md properly documents the vulnerability and prevention learned.

**Overall:** Good direction for security hardening. The two things I'd actually fix before merging are (1) the `os.homedir()` null safety in `index.ts`, and (2) making the Hermes cleanup more targeted so it doesn't risk touching unrelated YAML sections.

giauphan:
cc @jules

**What this PR does:** Migrates `CODEATLAS_API_KEY` from plaintext embedding in MCP config files (Hermes, Claude Desktop) to a secure `~/.codeatlas/.env` file with `0o600` permissions, then loads it globally via `dotenv`. Also updates sentinel.md and cleans up legacy embedded keys during migration.

### Issues

1. **Race condition on env file writes** — Two concurrent invocations of `setup_second_brain` could both read the existing `.env`, one overwrites, the other writes back its stale+modified content, silently dropping the first write's key. The comment even says "partial mitigate." If you want to be stricter, use a lock file or atomic temp-file + rename. Low urgency given this is a local dev tool.

2. **Hermes cleanup regex is fragile** — `[ \t]*CODEATLAS_API_KEY:.*(\r?\n|$)` will strip any line containing that pattern, which is fine, but the follow-up empty-env-block cleanup (`[ \t]*env:[ \t]*(\r?\n)(?![ \t]+[A-Za-z0-9_]+:)`) has an odd negative lookahead that only prevents stripping if there's *another* env entry on the next line. If `env:` was the last key in the block, it gets removed. This seems correct but the regex is non-obvious — a comment explaining intent would help. More importantly, the cleanup runs before checking whether `codeatlas:` exists at all, so it can strip unrelated `env:` blocks if the user happened to have a similar pattern nearby. Should guard cleanup behind detecting the `codeatlas:` section specifically.

3. **Claude Desktop `delete` doesn't preserve structure safely** — After deleting empty `env` objects, if either server's only remaining property was `env`, the entire server entry becomes `{ command: "...", args: ["..."] }` which is fine. But if some other property existed (unlikely), `Object.keys(...).length === 0` correctly cleans up. This is OK.

4. **Missing error handling for home directory** — `os.homedir()` can return `null` on some systems. `path.join(null, ".codeatlas")` produces `.codeatlas` (current dir), which writes without permission checks. The `getHomePath()` call already guards this, but `index.ts` uses `os.homedir()` directly without the same check. Should wrap in a try/catch or use `getHomePath()` consistently.

5. **Sentinel.md reordering** — The entries were reordered by date. Not a bug, but worth noting that chronologically sorting them would make the file easier to scan. Currently it goes 2025-02 → 2024-08 → 2025-02.

### Green flags

- File permissions (`0o700` for dir, `0o600` for env) are correct.
- The `override: false` in `dotenv.config()` is right — don't clobber already-set env vars.
- Cleaning up legacy `env: { CODEATLAS_API_KEY: ... }` references in existing configs prevents dangling secrets.
- Sentinel.md properly documents the vulnerability and prevention learned.

**Overall:** Good direction for security hardening. The two things I'd actually fix before merging are (1) the `os.homedir()` null safety in `index.ts`, and (2) making the Hermes cleanup more targeted so it doesn't risk touching unrelated YAML sections.

---

giauphan (commented):
cc @jules

I'm reviewing the code changes in this diff. Let me break down my analysis.

## 1. Correctness ✅ Looks solid

- The global env loading (`index.ts`) is defensive with proper `try/catch` and `override: false` to avoid clobbering local values.
- In `mcpServer.ts`, the regex replacement for updating/adding `CODEATLAS_API_KEY` in the .env file handles both append and replace scenarios correctly.
- The cleanup logic removes old embedded keys from Claude Desktop config by deleting them from the JSON objects before merging back.
- Hermes cleanup also strips empty `env:` blocks after removing keys.

**One minor concern:** The Claude Desktop cleanup uses `delete` which is fine for plain JS objects, but if `existing.mcpServers["codeatlas-genome"].env` has other properties being deleted via `delete`, those would be lost — however since we're only storing the key there now, this is correct behavior (we removed the key entirely).

## 2. Security ✅ Improved

This PR actually **fixes a real security issue** rather than introducing one:

- **Before:** API keys were written directly into Claude Desktop's `claude_desktop_config.json` and Hermes' YAML config as plaintext strings embedded in config files that might be committed to version control or readable by unintended users.
- **After:** Keys are stored in `~/.codeatlas/.env` with `0o600` permissions, and client configs no longer contain embedded credentials.

**Potential attack vector to consider:** The `fs.writeFileSync(envPath, envContent, { mode: 0o600 })` call — when the file already exists, `{ mode: 0o600 }` on some platforms may not preserve the intended permission if the filesystem ignores the `mode` flag on existing files. On Linux it works fine for new files but for overwrites, `chmod` explicitly afterward would be safer. Minor concern though.

## 3. Performance ⚠️ One item

In the Hermes cleanup block (line ~2301):
```typescript
if (cfg.includes("codeatlas:")) {
  fs.writeFileSync(hermesCfg, cfg); // Save cleanup
```

This unconditionally writes the entire Hermes config even when no actual change was made (e.g., no `CODEATLAS_API_KEY:` line existed to clean up). If someone has many MCP entries, this could cause unnecessary I/O. Not a correctness bug, just minor inefficiency. Should add a guard to only write if `cfg` actually changed.

## 4. Code Quality ⚠️ 

Several small nitpicks:

- **Redundant `else` clause:** After `if (!envContent.includes(...))` with an `else`, the code falls through naturally but it reads slightly awkwardly. Fine functionally.
- **Mixed error styles:** Some catch blocks log silently (good for config loading), while others push to `results`. Consistent.
- **`let mcpEntry = ...` followed by another assignment:** The first `const mcpEntry = ...` is moved below the new env-saving block, creating two sequential assignments. Readable enough.
- **Comment says "partial mitigate" for race conditions:** This is honest but worth noting — concurrent runs could still race. If this tool can run simultaneously from multiple sessions, consider locking around the .env write.

## 5. Maintainability ✅ Good

- The changes follow existing patterns in the codebase.
- Error handling is consistent (graceful degradation for optional config files).
- The sentinel doc update documents the original vulnerability well, making it easier for future devs to avoid repeating the same mistake.

---

**Verdict:** This is a well-executed security improvement. The main action items are:

1. **(Optional)** Add a guard before writing unchanged configs in the Hermes cleanup section
2. **(Optional)** Consider explicit `chmod 0o600` on the env file after overwrite for robustness across platforms
3. No critical issues found. Ship it.

giauphan:
cc @jules

I'm reviewing the code changes in this diff. Let me break down my analysis.

## 1. Correctness ✅ Looks solid

- The global env loading (`index.ts`) is defensive with proper `try/catch` and `override: false` to avoid clobbering local values.
- In `mcpServer.ts`, the regex replacement for updating/adding `CODEATLAS_API_KEY` in the .env file handles both append and replace scenarios correctly.
- The cleanup logic removes old embedded keys from Claude Desktop config by deleting them from the JSON objects before merging back.
- Hermes cleanup also strips empty `env:` blocks after removing keys.

**One minor concern:** The Claude Desktop cleanup uses `delete` which is fine for plain JS objects, but if `existing.mcpServers["codeatlas-genome"].env` has other properties being deleted via `delete`, those would be lost — however since we're only storing the key there now, this is correct behavior (we removed the key entirely).

## 2. Security ✅ Improved

This PR actually **fixes a real security issue** rather than introducing one:

- **Before:** API keys were written directly into Claude Desktop's `claude_desktop_config.json` and Hermes' YAML config as plaintext strings embedded in config files that might be committed to version control or readable by unintended users.
- **After:** Keys are stored in `~/.codeatlas/.env` with `0o600` permissions, and client configs no longer contain embedded credentials.

**Potential attack vector to consider:** The `fs.writeFileSync(envPath, envContent, { mode: 0o600 })` call — when the file already exists, `{ mode: 0o600 }` on some platforms may not preserve the intended permission if the filesystem ignores the `mode` flag on existing files. On Linux it works fine for new files but for overwrites, `chmod` explicitly afterward would be safer. Minor concern though.

## 3. Performance ⚠️ One item

In the Hermes cleanup block (line ~2301):
```typescript
if (cfg.includes("codeatlas:")) {
  fs.writeFileSync(hermesCfg, cfg); // Save cleanup
```

This unconditionally writes the entire Hermes config even when no actual change was made (e.g., no `CODEATLAS_API_KEY:` line existed to clean up). If someone has many MCP entries, this could cause unnecessary I/O. Not a correctness bug, just minor inefficiency. Should add a guard to only write if `cfg` actually changed.

## 4. Code Quality ⚠️ 

Several small nitpicks:

- **Redundant `else` clause:** After `if (!envContent.includes(...))` with an `else`, the code falls through naturally but it reads slightly awkwardly. Fine functionally.
- **Mixed error styles:** Some catch blocks log silently (good for config loading), while others push to `results`. Consistent.
- **`let mcpEntry = ...` followed by another assignment:** The first `const mcpEntry = ...` is moved below the new env-saving block, creating two sequential assignments. Readable enough.
- **Comment says "partial mitigate" for race conditions:** This is honest but worth noting — concurrent runs could still race. If this tool can run simultaneously from multiple sessions, consider locking around the .env write.

## 5. Maintainability ✅ Good

- The changes follow existing patterns in the codebase.
- Error handling is consistent (graceful degradation for optional config files).
- The sentinel doc update documents the original vulnerability well, making it easier for future devs to avoid repeating the same mistake.

---

**Verdict:** This is a well-executed security improvement. The main action items are:

1. **(Optional)** Add a guard before writing unchanged configs in the Hermes cleanup section
2. **(Optional)** Consider explicit `chmod 0o600` on the env file after overwrite for robustness across platforms
3. No critical issues found. Ship it.

---

PR #137, opened 2026-08-08
🔒 [security fix] Prevent Git Indirect Command Injection via Allowlist

giauphan:
🎯 **What:** Fixed an indirect command injection vulnerability in the `execGit` helper inside `mcpServer.ts`, which could be exploited through malicious or unexpected arguments bypassing the existing denylist.

⚠️ **Risk:** Although `shell: false` was used, an attacker passing unexpected arguments (e.g., negative numerical variables evaluated as flags, or global flags like `--exec-path` or `-c=...`) to `git` could trick git into executing arbitrary shell commands, leading to a severe system compromise. Pagers can also execute code if triggered unexpectedly.

🛡️ **Solution:** 
- Replaced the denylist with a robust allowlist to ensure only known-safe subcommands and arguments (`rev-parse`, `HEAD`, `--format=...`, etc.) are processed. 
- Prepended `--no-pager` as a global flag to eliminate any pager-related risks. 
- Prevented negative limits from dynamically forming unintended options by using `Math.max(1, limit)`.

---
*PR created automatically by Jules for task [5214672224064684317](https://jules.google.com/task/5214672224064684317) started by @giauphan*


giauphan (commented):
cc @jules

Looking at this diff — it replaces a git argument denylist with a strict allowlist, adds `--no-pager` protection, bounds the commits input, and removes dead code blocks. Let me trace through carefully.

**Security (the main intent of this PR)**

The allowlist approach is much stronger than the previous denylist. A few things to verify:

1. **The regex `/^\-\d+$/` is broad.** It matches any negative integer like `-99`, which gets passed as a git argument position where an allowed arg was expected. This covers the "negative numbers interpreted as flags" concern mentioned in the sentinel note. But — if there are other legitimate non-allowed string args used elsewhere in `execGit` calls that aren't listed in `allowedArgs` (e.g., branch names, commit hashes, file paths), those will get blocked. I need to check all callers of `execGit` to make sure every possible argument value is covered.

2. **`a.startsWith("--format=")` is good** for dynamic format strings, but you're only using `allowedArgs.has(a)` for exact string matches. For `rev-list`, the `HEAD...@{upstream}` syntax and `--left-right`, `--count` are all literal strings hardcoded in the Set — that's correct for this usage.

3. **`--no-pager` prepending is solid** and directly addresses the pager injection vector from the sentinel note. No issue here.

4. **Potential gap:** The allowlist uses exact equality (`Set.has`), so if someone passes `"--format=%H %s"` as two separate args, it won't match. You'd need the full string `--format=%H %s` as one element. Make sure callers construct the argv correctly with `["--format=%H %s"]` not `["--format=%H", "%s"]`. If the current code does pass it correctly, fine — just something to double-check.

**Correctness**

5. **`Math.max(1, Math.min(commits || 5, 20))`** — this prevents `commits` of `0` or `undefined` from producing `0` maxCommits (which would yield no results), and caps at 20. Clean change. No bug here.

6. **Dead code removal** at lines 2987-2993 — arrays declared but never used after this point were removed. That's cleanup, good.

**Performance**

7. **No regressions.** `Set.has` is O(1) per lookup, replacing what was essentially `array.filter + array.length > 0` which iterated all elements. Slight improvement.

**Code Quality / Maintainability**

8. **The sentinel.md update is good practice** — documenting the vulnerability class prevents repeating the same mistake.

9. **One minor concern:** the error message changed from "Forbidden git arguments detected" (plural) to "Forbidden git argument detected" (singular). Fine cosmetically but worth noting.

10. **The allowlist is implicit to this one function** but could become unmaintainable if more git commands are added. Consider keeping it near the caller sites or adding a comment mapping which commands use which allowed args.

**Verdict:** The core security fix is sound. The main risk area is ensuring *every* argument path flowing into `execGit` either hits the allowlist explicitly or goes through the `--format=` prefix check (for dynamic format strings). Check the actual call sites to confirm no legitimate arguments are being silently dropped.

giauphan:
cc @jules

Looking at this diff — it replaces a git argument denylist with a strict allowlist, adds `--no-pager` protection, bounds the commits input, and removes dead code blocks. Let me trace through carefully.

**Security (the main intent of this PR)**

The allowlist approach is much stronger than the previous denylist. A few things to verify:

1. **The regex `/^\-\d+$/` is broad.** It matches any negative integer like `-99`, which gets passed as a git argument position where an allowed arg was expected. This covers the "negative numbers interpreted as flags" concern mentioned in the sentinel note. But — if there are other legitimate non-allowed string args used elsewhere in `execGit` calls that aren't listed in `allowedArgs` (e.g., branch names, commit hashes, file paths), those will get blocked. I need to check all callers of `execGit` to make sure every possible argument value is covered.

2. **`a.startsWith("--format=")` is good** for dynamic format strings, but you're only using `allowedArgs.has(a)` for exact string matches. For `rev-list`, the `HEAD...@{upstream}` syntax and `--left-right`, `--count` are all literal strings hardcoded in the Set — that's correct for this usage.

3. **`--no-pager` prepending is solid** and directly addresses the pager injection vector from the sentinel note. No issue here.

4. **Potential gap:** The allowlist uses exact equality (`Set.has`), so if someone passes `"--format=%H %s"` as two separate args, it won't match. You'd need the full string `--format=%H %s` as one element. Make sure callers construct the argv correctly with `["--format=%H %s"]` not `["--format=%H", "%s"]`. If the current code does pass it correctly, fine — just something to double-check.

**Correctness**

5. **`Math.max(1, Math.min(commits || 5, 20))`** — this prevents `commits` of `0` or `undefined` from producing `0` maxCommits (which would yield no results), and caps at 20. Clean change. No bug here.

6. **Dead code removal** at lines 2987-2993 — arrays declared but never used after this point were removed. That's cleanup, good.

**Performance**

7. **No regressions.** `Set.has` is O(1) per lookup, replacing what was essentially `array.filter + array.length > 0` which iterated all elements. Slight improvement.

**Code Quality / Maintainability**

8. **The sentinel.md update is good practice** — documenting the vulnerability class prevents repeating the same mistake.

9. **One minor concern:** the error message changed from "Forbidden git arguments detected" (plural) to "Forbidden git argument detected" (singular). Fine cosmetically but worth noting.

10. **The allowlist is implicit to this one function** but could become unmaintainable if more git commands are added. Consider keeping it near the caller sites or adding a comment mapping which commands use which allowed args.

**Verdict:** The core security fix is sound. The main risk area is ensuring *every* argument path flowing into `execGit` either hits the allowlist explicitly or goes through the `--format=` prefix check (for dynamic format strings). Check the actual call sites to confirm no legitimate arguments are being silently dropped.

---

giauphan (commented):
cc @jules

This looks good. The main change replaces a flawed git argument denylist with a proper allowlist — a correct security fix. Here's my breakdown:

**Security ✅**
The old denylist had multiple bypass vectors: unbounded negative numbers (`git -123 exec evil`) could be interpreted as subcommands, `-cvalue` without space wouldn't match `-c`, and aliases weren't addressed. The allowlist fix closes all of these. `--no-pager` prevents a subtle info-leak vector too. Good work.

**Correctness ✅**
Adding `Math.max(1, ...)` properly handles the `commits=0` edge case (prevents empty output). Regex `/^\-\d+$/` correctly matches `-123` as a count while `a === "HEAD"` catches string arguments.

**Code quality ✅**
Cleanup of unused `modules/classes/functions` arrays is nice. Error messages are consistent.

**Minor notes (not blockers):**
- The allowlist is quite narrow — only `rev-parse`, `status`, `rev-list`, and `log --name-only` are permitted. If new tools later need different git subcommands, someone will need to remember to update this set. A brief comment noting this is a whitelist-by-design would help future maintainers.
- Each `execGit` call spawns a separate git process. If callers invoke it N times in a loop (e.g., one per file), that's N subprocesses. Worth being aware of, but unlikely to be a practical concern at current scale.

**Verdict:** LGTM. This is a legitimate security improvement with no introduced bugs.

giauphan:
cc @jules

This looks good. The main change replaces a flawed git argument denylist with a proper allowlist — a correct security fix. Here's my breakdown:

**Security ✅**
The old denylist had multiple bypass vectors: unbounded negative numbers (`git -123 exec evil`) could be interpreted as subcommands, `-cvalue` without space wouldn't match `-c`, and aliases weren't addressed. The allowlist fix closes all of these. `--no-pager` prevents a subtle info-leak vector too. Good work.

**Correctness ✅**
Adding `Math.max(1, ...)` properly handles the `commits=0` edge case (prevents empty output). Regex `/^\-\d+$/` correctly matches `-123` as a count while `a === "HEAD"` catches string arguments.

**Code quality ✅**
Cleanup of unused `modules/classes/functions` arrays is nice. Error messages are consistent.

**Minor notes (not blockers):**
- The allowlist is quite narrow — only `rev-parse`, `status`, `rev-list`, and `log --name-only` are permitted. If new tools later need different git subcommands, someone will need to remember to update this set. A brief comment noting this is a whitelist-by-design would help future maintainers.
- Each `execGit` call spawns a separate git process. If callers invoke it N times in a loop (e.g., one per file), that's N subprocesses. Worth being aware of, but unlikely to be a practical concern at current scale.

**Verdict:** LGTM. This is a legitimate security improvement with no introduced bugs.

---

giauphan (commented):
cc @jules

Here's my review:

**Security — Strong improvement.**

The old denylist approach was genuinely fragile. The new allowlist is much better, but I have one concern:

```ts
const isSafe = args.every(a => allowedArgs.has(a) || /^\-\d+$/.test(a) || a.startsWith("--format="));
```

The regex `^\-\d+$` accepts negative numbers like `-1`, `-2`, etc. These get passed to git as-is. While git won't execute them as flags (they're consumed as positional arguments by commands like `rev-list -n`), the regex also matches things like just `-5` which some git commands might interpret differently depending on context. This is low risk but worth being intentional about — consider binding these per-command instead of globally.

Also, `"--no-pager"` is prepended correctly. Good call.

**Correctness — One bug.**

Line 3 changes `Math.min(commits || 5, 20)` to `Math.max(1, Math.min(commits || 5, 20))`. This prevents `commits=0` from becoming 0, but if the caller passes `commits=0` they might *intentionally* want no results. However, given the downstream use in `rev-list --count`, passing 0 commits would produce empty output anyway, so enforcing a minimum of 1 is probably fine. Just be aware it's a behavior change.

**Code Quality — Good cleanup.**

The unused module/class/function variable declarations removed at line ~2984 were dead code. Clean removal, no issue there.

**Verdict:** Solid PR. The security hardening (allowlist + `--no-pager`) is a real improvement over the previous denylist. My only note is that the per-command argument policy could be tightened further down the road, but what's here is correct and safe enough for now.

giauphan:
cc @jules

Here's my review:

**Security — Strong improvement.**

The old denylist approach was genuinely fragile. The new allowlist is much better, but I have one concern:

```ts
const isSafe = args.every(a => allowedArgs.has(a) || /^\-\d+$/.test(a) || a.startsWith("--format="));
```

The regex `^\-\d+$` accepts negative numbers like `-1`, `-2`, etc. These get passed to git as-is. While git won't execute them as flags (they're consumed as positional arguments by commands like `rev-list -n`), the regex also matches things like just `-5` which some git commands might interpret differently depending on context. This is low risk but worth being intentional about — consider binding these per-command instead of globally.

Also, `"--no-pager"` is prepended correctly. Good call.

**Correctness — One bug.**

Line 3 changes `Math.min(commits || 5, 20)` to `Math.max(1, Math.min(commits || 5, 20))`. This prevents `commits=0` from becoming 0, but if the caller passes `commits=0` they might *intentionally* want no results. However, given the downstream use in `rev-list --count`, passing 0 commits would produce empty output anyway, so enforcing a minimum of 1 is probably fine. Just be aware it's a behavior change.

**Code Quality — Good cleanup.**

The unused module/class/function variable declarations removed at line ~2984 were dead code. Clean removal, no issue there.

**Verdict:** Solid PR. The security hardening (allowlist + `--no-pager`) is a real improvement over the previous denylist. My only note is that the per-command argument policy could be tightened further down the road, but what's here is correct and safe enough for now.

---

giauphan (commented):
cc @jules

I've analyzed the diff and found a real security regression in the new allowlist logic.

**Bug: `--count` is allowed as an arg but `rev-list --left-right --count "HEAD...@{upstream}"` will break.** The allowlist includes `"--left-right", "--count", "HEAD...@{upstream}"` — this works if args are passed individually like `["rev-list", "--left-right", "--count", "HEAD...@{upstream}"]`. That's fine since each string hits the allowlist via `.has()`. So that's actually correct.

The real issue is with **dynamic arguments from git commands**. Looking at how `execGit` is called throughout the tool, some invocations likely pass dynamic values (branch names, file paths, commit hashes) that won't match the allowlist. For example, `git log --name-only <branch>` would fail because `<branch>` isn't in the allowlist. Similarly, any command passing a filename or revision ref that isn't literally `"HEAD...@{upstream}"` or `"HEAD"` will throw a false positive. This will break functionality.

**Security improvement is solid** — switching from denylist to strict allowlist and prepending `--no-pager` are both correct fixes for Git argument injection. The regex `/^\-\d+$/` blocks negative numbers (which could become flags), which is good.

**Code quality:** The deleted Python code block in the diff appears to be leftover/stray code removal — that's fine cleanup.

**Minor concern:** The allowlist contains `"HEAD...@{upstream}"` as a literal arg match. If the actual upstream lookup passes something like `"origin/main"` or uses a variable branch name, it'll fail. Consider whether the downstream callers always pass exactly that string.

**Verdict:** Security fix is correct in principle, but the allowlist may be too narrow and will cause runtime failures on any git call passing dynamic refs, filenames, or other non-literal arguments. Test thoroughly across all callers of `execGit`.

giauphan:
cc @jules

I've analyzed the diff and found a real security regression in the new allowlist logic.

**Bug: `--count` is allowed as an arg but `rev-list --left-right --count "HEAD...@{upstream}"` will break.** The allowlist includes `"--left-right", "--count", "HEAD...@{upstream}"` — this works if args are passed individually like `["rev-list", "--left-right", "--count", "HEAD...@{upstream}"]`. That's fine since each string hits the allowlist via `.has()`. So that's actually correct.

The real issue is with **dynamic arguments from git commands**. Looking at how `execGit` is called throughout the tool, some invocations likely pass dynamic values (branch names, file paths, commit hashes) that won't match the allowlist. For example, `git log --name-only <branch>` would fail because `<branch>` isn't in the allowlist. Similarly, any command passing a filename or revision ref that isn't literally `"HEAD...@{upstream}"` or `"HEAD"` will throw a false positive. This will break functionality.

**Security improvement is solid** — switching from denylist to strict allowlist and prepending `--no-pager` are both correct fixes for Git argument injection. The regex `/^\-\d+$/` blocks negative numbers (which could become flags), which is good.

**Code quality:** The deleted Python code block in the diff appears to be leftover/stray code removal — that's fine cleanup.

**Minor concern:** The allowlist contains `"HEAD...@{upstream}"` as a literal arg match. If the actual upstream lookup passes something like `"origin/main"` or uses a variable branch name, it'll fail. Consider whether the downstream callers always pass exactly that string.

**Verdict:** Security fix is correct in principle, but the allowlist may be too narrow and will cause runtime failures on any git call passing dynamic refs, filenames, or other non-literal arguments. Test thoroughly across all callers of `execGit`.

---

giauphan:
@jules Closing as duplicate. #140 fixes the same git-args allowlist injection in the same file with a more complete implementation (adds `--no-pager` global flag, `Math.max/Math.min` boundary clamp, and a 5 MB `maxBuffer` cap for DoS mitigation).

#140 is currently `MERGEABLE`. Please port any missing pieces (e.g. additional allowlist entries) into #140 if needed, and coordinate with the author there.

PR #143, opened 2026-08-10
🛡️ Sentinel: [MEDIUM] Prevent String Replacement Injection in API Key configuration

giauphan:
What: Fixed a string replacement injection vulnerability where `String.prototype.replace()` evaluated special patterns in the user-supplied API key instead of treating it as a literal string.
Risk: If the API key contained patterns like `$&` or `$1`, it would be evaluated as a token rather than a string when appended to `.env`, potentially corrupting the `.env` file configuration and leaking unintended environment values.
Solution: Used an arrow function as the replacement parameter in `String.prototype.replace()` instead of a string template to prevent interpolation of special patterns.

---
*PR created automatically by Jules for task [15265757483805258364](https://jules.google.com/task/15265757483805258364) started by @giauphan*


giauphan:
@jules Closing as duplicate. The identical fix (replacer function for CODEATLAS_API_KEY line in mcpServer.ts) was already merged via PR #147 into main. No action needed on this PR.

PR #181, opened 2026-08-29
⚡ Bolt: [performance improvement] Optimize intermediate array allocations in MCP tools

giauphan:
💡 What: Replaced chained `.filter().slice().map()` and `.toLowerCase().includes()` inside `sync_skills_inventory` with a single `for...of` loop and a precompiled regular expression (`new RegExp`). Similarly replaced `.slice().map().filter()` inside `code_search` with efficient manual `for` loops.
🎯 Why: Chained array methods allocate temporary arrays and strings for each stage of the chain, causing massive Garbage Collection (GC) overhead when processing large codebases or datasets.
📊 Impact: Considerably reduces memory consumption and GC spikes by avoiding unnecessary O(N) array traversals and intermediate allocations during context extraction and inventory processing.
🔬 Measurement: Verify by running `pnpm run test` and `pnpm lint`. The optimization replaces identical logic so the results remain identical while avoiding memory spikes.

---
*PR created automatically by Jules for task [3225468169628676403](https://jules.google.com/task/3225468169628676403) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -78,11 +78,9 @@
 ## 2026-08-27 - [Performance improvement] Optimized multiple array allocations by avoiding `.slice().filter().slice()`
 **Learning:** Chaining array methods like `ls.slice().filter().slice()` creates multiple intermediate arrays, causing significant GC overhead, especially when parsing large text blocks (like git commit logs).
 **Action:** When extracting a small subset of elements from a large array based on a condition, avoid chained array methods. Instead, use a single `for` loop, conditionally push elements, and break early when the limit is reached.
-=======
 ## 2024-05-24 - [Performance improvement] Optimized O(N) chained array .filter() operations across multiple metrics
 **Learning:** Chaining array methods (e.g. `.filter(n => n.filePath).length`) and mapping over arrays separately to calculate simple metrics like counts or orphan entities wastes execution time and creates large intermediate array memory allocations.
 **Action:** When computing multiple simple metrics across a graph or dataset (e.g., node typing, relationships, existence of file paths), combine all checks into a single `for...of` loop with simple accumulators (`count++`, `Map.set`) to collapse multiple O(N) array loops into a single O(N) pass.
->>>>>>> 819a139 (refactor: remove noisy bolt comment)
 ## 2026-08-31 - [Performance improvement] Optimized O(N) array filtering and memory allocations when parsing data
 **Learning:** Chaining array methods like `.filter().map().slice()` creates multiple intermediate array and string allocations, increasing Garbage Collection overhead. Specifically, using `.slice().map(l => l.trim()).filter(Boolean)` iteratively on file lines causes massive GC spikes in tools like `code_search`.
 **Action:** When extracting a subset of elements (like code context blocks) or filtering items dynamically, avoid array method chaining. Instead, use a single `for` or `for...of` loop with early breaks, and push matching, pre-processed elements into a single result array to prevent unnecessary allocations.
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -83,3 +83,6 @@
 **Learning:** Chaining array methods (e.g. `.filter(n => n.filePath).length`) and mapping over arrays separately to calculate simple metrics like counts or orphan entities wastes execution time and creates large intermediate array memory allocations.
 **Action:** When computing multiple simple metrics across a graph or dataset (e.g., node typing, relationships, existence of file paths), combine all checks into a single `for...of` loop with simple accumulators (`count++`, `Map.set`) to collapse multiple O(N) array loops into a single O(N) pass.
 >>>>>>> 819a139 (refactor: remove noisy bolt comment)
+## 2026-08-31 - [Performance improvement] Optimized O(N) array filtering and memory allocations when parsing data
+**Learning:** Chaining array methods like `.filter().map().slice()` creates multiple intermediate array and string allocations, increasing Garbage Collection overhead. Specifically, using `.slice().map(l => l.trim()).filter(Boolean)` iteratively on file lines causes massive GC spikes in tools like `code_search`.
+**Action:** When extracting a subset of elements (like code context blocks) or filtering items dynamically, avoid array method chaining. Instead, use a single `for` or `for...of` loop with early breaks, and push matching, pre-processed elements into a single result array to prevent unnecessary allocations.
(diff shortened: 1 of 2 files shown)


giauphan:
cc @jules

## Code Review

### `.jules/bolt.md`

`.jules/bolt.md` -> Line 86: Leftover merge conflict marker (`>>>>>>>`) from a previous commit is still present in the file. This should be cleaned up before merging.

---

**Suggestions loop (lines ~1107–1114)**

The refactor is correct and achieves the stated goal (single-pass, no intermediate arrays). No issues.

**Git status parsing (lines ~2303–2311)**

The refactor is correct and more readable than the original one-liner. However:

`src/presentation/mcpServer.ts` -> Line ~2311: The status code check `s.includes("M")` is fragile. `git status --porcelain` uses a two-character field where index status is column 0 and worktree status is column 1 (e.g., `MM`, ` M`, `AM`). Using `includes("M")` on a 2-char string is functionally equivalent to checking either column, but it will also match `MM` twice (harmless since it pushes to the same array) and could in theory match unexpected states. Prefer explicit checks like `s[0] === "M" || s[1] === "M"` to make intent clear. Same applies to `"A"` and `"D"`.

`src/presentation/mcpServer.ts` -> Line ~2311: An `A` status in column 0 followed by `D` in column 1 (i.e., `AD` — added then deleted) would push the same file into both `add` and `del`. The original code had the same issue, but since this is a refactor touching this logic, it's worth flagging.

---

### Overall

The performance intent is sound — both changes eliminate chained array methods and intermediate allocations. The merge conflict artifact in the markdown file is the most concrete issue to fix before merging. The status-parsing logic is a pre-existing bug carried through, not introduced here, but worth addressing.

---

giauphan:
cc @jules

## Code Review

### `.jules/bolt.md`
No issues. Conflict markers removed cleanly, new learning entry is well-formed.

---

**File: `src/presentation/mcpServer.ts` -> Line ~2309-2316: Logic bug in git status parsing**

The new character-by-character status check introduces a regression. Git `--porcelain` format uses a two-character XY status code where `X` = index status, `Y` = worktree status. The old code used `s.includes("M")` which (while imprecise) caught both positions. The new code checks `s[0]` and `s[1]`, which is correct in principle — but it will incorrectly flag files as both modified *and* added/deleted when the XY combo mixes statuses (e.g., `AM` → added in index, modified in worktree would push to both `add` and `mod`). Whether that's desired behavior depends on intent, but it's a semantic change that should be deliberate, not accidental.

More critically: `s.includes("D")` in the old code correctly caught `" D"` (deleted in worktree, no index change). The new `s[0] === "D"` check misses `" D"` since `s[0]` would be `" "`. Verify that `s[1] === "D"` is actually reached for that case — it is, so deletions are fine, but the double-push risk above still stands.

**File: `src/presentation/mcpServer.ts` -> Line ~1107-1114: Minor — acceptable tradeoff**

The refactor from `.filter().map().slice()` to an early-exit loop is correct and improves performance. No issues.

---

### Summary

The main concern is the git status parsing logic change. The `.includes()` → index-check refactor is a **behavioral change** (double-counting files with mixed XY statuses), not just a performance refactor. This should be validated against porcelain output with staged+unstaged changes on the same file (e.g., `MM` case). Everything else looks good.

---

giauphan:
cc @jules

## Code Review

**Line ~2316 (git status parsing):** The comment says "prevent duplicate pushes to the same array (e.g. MM)" but the logic doesn't actually prevent it. For a status of `MM`, both `s[0] === "M"` and `s[1] === "M"` are true, so `mod.push(f)` is called twice. The original `s.includes("M")` had the same bug, but the comment falsely claims this is fixed.

`src/presentation/mcpServer.ts` -> Line ~2318: Comment is misleading — `MM` status still causes duplicate push to `mod[]`. Either fix it with a single check (`if (s.includes("M")) mod.push(f)`) or use `s[0] === "M" || s[1] === "M"` but remove the incorrect claim about preventing duplicates.

**Line ~2316 (git status parsing):** The inline comment says "do not trim entirely to preserve space padding" but then calls `.trim()` on `f` (`line.substring(3).trim()`). Trimming `f` is correct for filenames, and `substring(0,2)` correctly reads the two status chars without trimming, so the code is fine — but the comment is confusing and slightly inaccurate.

`src/presentation/mcpServer.ts` -> Line ~2310: Comment could mislead future readers. The actual intent is "don't trim the full line before reading status columns 0–1" — worth rewording for clarity.

---

### Everything else looks good

- The `suggestions` loop refactor (lines 1107–1119) is a correct and clean optimization — avoids three intermediate arrays, early-breaks at 10, and produces identical output.
- The `.jules/bolt.md` change is a docs-only cleanup removing leftover git conflict markers. No issues.

---

giauphan:
cc @jules

## Code Review

**Line ~1107–1114: Suggestions loop — looks good overall**

The refactor from `.filter().map().slice(0, 10)` to a `for` loop with early break is a clean performance improvement. No issues.

---

**Line ~2303–2322: Git status parsing — logic bug**

`src/presentation/mcpServer.ts` -> Line ~2314–2322: The new index-based status checks (`s[0] === "M"`, `s[1] === "M"`, etc.) do **not** faithfully replicate the original `s.includes("M")` behaviour.

`git status --porcelain` uses a two-character XY format where `X` is the index (staging area) status and `Y` is the worktree status. The original `s.includes("M")` would match `M` in *either* position, which the new code also does — so that part is equivalent.

However, the original also matched `s.includes("A")` and `s.includes("D")`, which would catch status codes like `"AD"` (added in index, deleted in worktree) pushing the file into **both** `add` and `del`. The new code does the same via separate if-blocks, so that's preserved.

The real edge-case issue: `git status --porcelain` can output **rename entries** like `R  old-name -> new-name` (two-file entries on one line). The filename after column 3 in that case is `"old-name -> new-name"`, which the code would store verbatim. The original code had the same problem, but since this is a refactor it's worth noting.

Also minor: `s[0]` and `s[1]` on an empty string would return `undefined`, not throw — but the `!line.trim()` guard above ensures `line` is non-empty, so this is safe.

`src/presentation/mcpServer.ts` -> Line ~2322: `f = line.substring(3).trim()` — adding `.trim()` here is **a behaviour change**. The original code used `line.substring(3)` on an already-trimmed line, which would trim leading whitespace from the whole line (including the status columns). The new code correctly preserves columns 0–1 by not trimming the full line, then trims only the filename part. This is actually a **bug fix** — the old `x.trim()` on the full line would corrupt the XY prefix for lines with leading spaces. The new approach is more correct.

---

**Overall**

The changes are reasonable refactors for performance and readability. The logic is largely equivalent to the original. The main thing worth double-checking is whether rename (`R`/`C`) entries from `git status --porcelain` need to be handled — but that was pre-existing, not introduced here. No security or maintainability concerns. No hardcoded secrets. Looks good to merge.

---

giauphan:
Superseded by canonical merged PR #199

PR #192, opened 2026-09-04
⚡ Bolt: Replace O(N log N) sorting with O(N) bucket collection

giauphan:
💡 What: Replaced the `Array.prototype.sort()` + `Array.prototype.indexOf()` node prioritization logic in `parser.ts` and `mcpServer.ts` with an O(N) bucket-collection strategy.
🎯 Why: Using `indexOf` inside an O(N log N) sort comparator creates a compounding performance bottleneck, especially when processing graphs containing tens or hundreds of thousands of nodes.
📊 Impact: Expected to significantly reduce processing latency during initial graph loading and truncation operations, dropping the time complexity for node prioritization from O(N log N) to O(N).
🔬 Measurement: Verified through ad-hoc microbenchmarking and by ensuring all 98 unit tests continue to pass correctly, maintaining sorting stability and correctness.

---
*PR created automatically by Jules for task [16312005699897506194](https://jules.google.com/task/16312005699897506194) started by @giauphan*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -83,3 +83,7 @@
 **Learning:** Chaining array methods (e.g. `.filter(n => n.filePath).length`) and mapping over arrays separately to calculate simple metrics like counts or orphan entities wastes execution time and creates large intermediate array memory allocations.
 **Action:** When computing multiple simple metrics across a graph or dataset (e.g., node typing, relationships, existence of file paths), combine all checks into a single `for...of` loop with simple accumulators (`count++`, `Map.set`) to collapse multiple O(N) array loops into a single O(N) pass.
 >>>>>>> 819a139 (refactor: remove noisy bolt comment)
+
+## 2024-05-20 - Array.prototype.sort with indexOf Bottleneck
+**Learning:** Using `Array.prototype.sort()` combined with `Array.prototype.indexOf()` inside the comparator for prioritizing large collections of elements creates an O(N log N) performance bottleneck. The constant $O(K)$ lookup overhead of `indexOf` inside the $O(N \log N)$ sorting loop exacerbates the slowness, particularly in Node.js event loops when sorting thousands of items.
+**Action:** When categorizing or prioritizing elements into a predefined set of buckets, replace sorting with an O(N) bucket-collection strategy. Linearly iterate over the collection, allocate elements to their respective arrays in a dictionary, and concatenate the buckets to preserve stability and significantly improve execution speed. Always use prototype-safe key lookups (`Object.hasOwn`) when mapping dynamic properties to bucket dictionary keys.
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,8 @@
       "bin": {
         "codeatlas": "dist/index.js",
         "codeatlas-enterprise": "dist/index.js",
-        "codeatlas-mcp": "dist/index.js"
+        "codeatlas-mcp": "dist/index.js",
+        "codeatlas-setup-claude": "dist/src/cli/setup-claude.js"
       },
       "devDependencies": {
         "@semantic-release/changelog": "7.0.0",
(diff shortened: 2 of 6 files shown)


giauphan:
cc @jules

I now have a full picture of both the PR diff and the current repo state. Let me compile my findings.

---

## PR Review: Performance Optimizations for CodeAtlas Analyzer

### Critical Issues

**1. `.orig` backup files committed to repository**
- `src/analyzer/parser.ts.orig` and `src/presentation/mcpServer.ts.orig` (3406 lines!) are merge/backup artifacts that should never be in version control. These bloat the repo and can cause confusion about which file is canonical.

**2. Prototype pollution vulnerability in `src/presentation/mcpServer.ts`**
```typescript
// Line ~547 and ~345 (in generate_system_flow)
if (buckets[node.type]) {
    buckets[node.type].push(node);
} else {
    buckets['other'].push(node);
}
```
Direct property access `buckets[node.type]` can reach inherited properties like `__proto__`, `constructor`, or `toString`. If `node.type` is `"constructor"`, `buckets["constructor"]` returns the Object constructor (truthy), and `.push()` on it throws or behaves unexpectedly. The `parser.ts` version correctly uses `Object.hasOwn(buckets, node.type)` — this should be matched here.

### Type Safety Issues

**3. Unnecessary `as string` type assertion in `src/analyzer/parser.ts`**
```typescript
node.type !== 'other' as string
```
This should be `node.type !== 'other'`. The `as string` is redundant (string literals are already strings). Furthermore, this entire `'other'` check is logically unnecessary — `Object.hasOwn(buckets, 'other')` is true, and nodes with type `'other'` would still be pushed to `buckets['other']` even without this guard.

**4. Double type assertion bypassing type checking in `src/analyzer/parser.ts`**
```typescript
reversedClasses as unknown as ClassReference[]
```
The `as unknown as` pattern indicates `reversedClasses` and `ClassReference[]` have incompatible types. This silently bypasses TypeScript's type checker and could hide real type mismatches. The underlying type issue should be resolved instead.

**5. `as any` casts in `src/presentation/mcpServer.ts`**
In `sync_dreams` tool:
```typescript
const typed = d as any;
```
and later `d as any`. These defeat TypeScript's type safety. Should use `Record<string, unknown>` or proper interfaces.

### Error Handling

**6. Silent error swallowing in `getDirContents`**
```typescript
catch {
    contents = null;
}
```
Empty catch blocks hide all errors (permissions issues, race conditions, etc.) with no logging. At minimum should call `this.reportError()` or `this.reportWarning()`.

**7. `analyzeFileIncremental` assumes `analyzeFile` returns boolean**
The new code does `const success = this.analyzeFile(absPath)` and checks `if (success)`, but if `analyzeFile` returns `void` or `undefined`, files will never be tracked in `allFiles`, breaking incremental analysis. The original code didn't check the return value.

### Documentation/Consistency

**8. Inconsistent truncation default in `get_project_structure`**
- Schema documents: `default: 100`
- Code implements: `const maxResults = limit || 500;`

### Repo-level Issues (unrelated to PR diff but blocking)

**9. Missing methods on `GitHubClient`**
`src/index.ts` (line 84) and `src/webhook/handler.ts` (line 57) both call `gitHub.fetchDiff()`, and the handler also calls `gitHub.postReview()` (line 169) and `gitHub.postInlineComments()` (line 178) — none of these exist on `GitHubClient`. This would cause `TypeError: gitHub.fetchDiff is not a function` at runtime.

---

**Summary:** The performance optimizations (bucket sort replacing O(N log N) sort, single-pass link traversal, precomputed node degrees) are sound and well-implemented in `parser.ts`. However, the `.orig` files should be removed, the prototype pollution in `mcpServer.ts` bucket logic should be fixed to match `parser.ts`'s `Object.hasOwn` approach, and the type assertions should be cleaned up. The missing `fetchDiff`/`postReview`/`postInlineComments` methods in `GitHubClient` are blocking bugs for the webhook functionality.

---

giauphan:
Superseded by canonical merged PR #206 (feat/bucket-sort-opt).

PR #205, opened 2026-09-09
🛡️ Sentinel: [CRITICAL] Fix SSRF protocol validation in API URL

giauphan:
🚨 Severity: CRITICAL
💡 Vulnerability: Server-Side Request Forgery (SSRF) risk via loosely validated API URL environment variable that previously allowed plaintext `http:` connections to external domains.
🎯 Impact: An attacker could potentially reconfigure or intercept the API URL to transmit sensitive application data (such as internal memory payloads or API keys) over unencrypted networks or exploit internal services if network topology allowed.
🔧 Fix: Enforced a strict protocol allowlist in `getApiUrl()` within `src/utils/envUtils.ts`. Specifically, `https:` is globally permitted, whereas `http:` is heavily restricted strictly to safe local hostnames (`localhost` or `127.0.0.1`). Unsafe configurations trigger sanitized errors that don't leak hostname details.
✅ Verification: Ran `npm run test`, successfully executing 142 tests including newly added boundary tests in `src/utils/envUtils.test.ts` for strict hostname validation on the `http:` protocol.

---
*PR created automatically by Jules for task [4490404917688878477](https://jules.google.com/task/4490404917688878477) started by @giauphan*

PR #209, opened 2026-09-10
🛡️ Sentinel: [CRITICAL] Prevent SSRF bypass by enforcing getApiUrl() usage

giauphan:
## Summary
- Replaces direct `process.env.CODEATLAS_API_URL` reads in core TypeScript network code with the hardened `getApiUrl()` helper
- Ensures attacker-controlled hosts (like AWS IMDS or random websites) cannot bypass the intended loopback/HTTPS restrictions

This addresses a bypass introduced and found post-merge of #205.

## Test plan
- Run tests and ensure cloud endpoints use the restrictive URL policy
- Ensure CI pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/presentation/mcpServer.ts
+++ b/src/presentation/mcpServer.ts
@@ -1,3 +1,4 @@
+import { getApiUrl } from "../utils/envUtils.js";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import * as fs from "fs";
@@ -947,8 +948,7 @@ export function registerTools(server: McpServer) {
       const auth = await checkAuth();
       await logActivity(auth, "search_genome", { query: query.substring(0, 100), project, limit });
       try {
-        const serverUrl = process.env.CODEATLAS_API_URL;
-        if (!serverUrl) throw new Error("CODEATLAS_API_URL not set");
+        const serverUrl = getApiUrl();
         const apiKey = process.env.CODEATLAS_API_KEY;
         if (!apiKey) throw new Error("CODEATLAS_API_KEY not set");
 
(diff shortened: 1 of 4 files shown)


giauphan (commented):
cc @jules

## Code Review

The PR extracts repeated `process.env.CODEATLAS_API_URL` lookups into a shared `getApiUrl()` utility function. This is a clean refactor that reduces duplication across 8 call sites. Good change overall, but there are a few issues:

### Issues

**File: src/utils/envUtils.ts -> Missing file:** The diff imports `getApiUrl` from `../utils/envUtils.js` but the implementation of this utility file is not included in the PR. Without seeing the implementation, we can't verify:
- Whether it throws with a clear error message (like the original `"CODEATLAS_API_URL not set"` did)
- Whether the return type is `string` (not `string | undefined`) — this is critical since all call sites now use the return value directly without null checks (e.g., `new URL(getApiUrl())` would throw a cryptic URL parse error on `undefined`)
- Whether it avoids the `any` type

**File: src/presentation/mcpServer.ts -> Line 2609:** The inconsistent indentation on the `fetch` call that was already present wasn't fixed:
```typescript
        const apiUrl = getApiUrl();
                const resp = await fetch(`${apiUrl}/api/genome/search?limit=1`, {
```
Minor, but since you're touching this code, it's a good opportunity to clean it up.

**File: src/presentation/mcpServer.ts -> Lines 948, 982, 1017, 1054:** The `CODEATLAS_API_KEY` check is still duplicated inline at every call site with the same pattern:
```typescript
const apiKey = process.env.CODEATLAS_API_KEY;
if (!apiKey) throw new Error("CODEATLAS_API_KEY not set");
```
Consider also extracting a `getApiKey()` utility for consistency — same rationale as this PR.

### Summary

The refactor direction is correct and improves maintainability. Please ensure `getApiUrl()` is included in the PR, returns `string` (not `string | undefined`), throws a descriptive error when the env var is missing, and avoids `any` types.

src/presentation/mcpServer.ts
```diff
@@ -2609,8 +2606,7 @@ def register(ctx):
```
giauphan, line 2606: src/presentation/mcpServer.ts -> Line 2609:** The inconsistent indentation on the `fetch` call that was already present wasn't fixed:

src/presentation/mcpServer.ts
```diff
@@ -1,3 +1,4 @@
+import { getApiUrl } from "../utils/envUtils.js";
```
giauphan, line 1: src/presentation/mcpServer.ts -> Lines 948, 982, 1017, 1054:** The `CODEATLAS_API_KEY` check is still duplicated inline at every call site with the same pattern:

PR #221, opened 2026-09-12
🛡️ Sentinel: [CRITICAL] Fix indirect command injection via execSync

giauphan:
🚨 Severity: CRITICAL
💡 Vulnerability: The use of `execSync` for executing node scripts implicitly relies on the system shell or the raw `node` string literal from the environment PATH. This invites command injection risks if user inputs or paths are unexpectedly manipulated.
🎯 Impact: An attacker who manages to tamper with PATH or inject input could achieve arbitrary shell command execution via shell escaping or by replacing the executable logic.
🔧 Fix: Replaced `execSync` instances in script execution logic across `commands.ts`, `setupClaude.ts`, `setup-claude.js`, and `validate-hooks.js` with `spawnSync`, specifically ensuring `shell: false`. Where Node.js scripts are run, switched to `process.execPath` to guarantee the current active binary is utilized.
✅ Verification: Ran `pnpm install`, `pnpm build`, `pnpm lint`, and `pnpm test`. All tests passed, meaning correct setup hooks operations are preserved safely.

---
*PR created automatically by Jules for task [17007448498509729234](https://jules.google.com/task/17007448498509729234) started by @giauphan*


giauphan:
cc @jules

This PR replaces `execSync` (shell-based) with `spawnSync(..., shell: false)` + `process.execPath` — the right hardening move, and the translation is mostly faithful. Argument arrays are correctly formed, `input:` properly replaces the `echo | bash` pipe, env handling is preserved, and no `any` types were introduced. A few minor points:

- File: src/cli/setupClaude.ts -> Line 82: On Windows, npm global bins are `.cmd` shims, and Node refuses to spawn `.cmd`/`.bat` with `shell: false` (EINVAL). The version check will now always fail on Windows and silently drop to the npx fallback — graceful, but a behavior regression worth knowing about.
- File: src/cli/setupClaude.ts -> Line 85: `(result.stdout as string)` casts away the `string | null` type; if stdout were ever null, `.trim()` throws a TypeError that the catch would mislabel as "CLI not found". A `result.stdout?.trim() ?? ''` or null check is cleaner than the cast.
- File: src/cli/commands.ts -> Line 733 (also setup-claude.js:33): The `if (result.error) throw ...; if (result.status !== 0) throw ...` guard pair is repeated 4 times across files — a tiny shared helper would avoid drift, but it's a nit.

Otherwise this is a solid change — no injection surface left, error paths behave equivalently to the old catch/throw flow, and the sentinel.md note matches the code change. Nothing blocking.

---

giauphan:
cc @jules

The PR replaces `execSync` with `spawnSync`/argument arrays to eliminate shell injection — the core change is correct and behavior-preserving. I verified each call site: `process.execPath` for node scripts, `shell: false`, the `win32` shell exception for `.cmd` wrappers, and the new `checkSpawnResult` helper (whose `string | Buffer` param accepts both encoded and inherit-mode results). A few minor things worth noting:

- File: src/utils/processUtils.ts -> Line 4: `result.status` is `null` when the child is killed by a signal (e.g. timeout), producing the confusing message "Process exited with status null". Consider `if (result.status !== 0) throw new Error(`Process exited with status ${result.status}${result.signal ? ` (signal ${result.signal})` : ""}`)`.
- File: src/cli/commands.ts -> Line 664: mid-file `import` works (ESM hoists) but is unconventional; move it to the top import block with the other imports.
- File: scripts/validate-hooks.js -> Line 105: unlike `test-brain-context` (line 136), this check doesn't guard `result.error` before reading `result.stderr` — on a spawn failure (e.g. timeout) `status` is `null` and output may be empty, so it still returns `false`, which is the right outcome, but the inconsistency is worth aligning.

No correctness, security, or performance issues introduced; everything else looks good.

---

giauphan:
cc @jules

The workspace doesn't contain the PR's files (this is the review-bot repo, not the target codebase), so I reviewed the diff directly, deriving line numbers from the hunk headers.

**Overall: good security hardening — no blocking issues.** The `execSync` → `spawnSync` conversions are correct (arg arrays, `shell: false`, `process.execPath`, `input` instead of an `echo |` pipe), and `checkSpawnResult` checks `error` before `status`, which is the right order. Three minor nits:

1. File: src/cli/commands.ts -> Line 664: `import` is placed mid-file (inside the CLI router section). Valid ESM (hoisted) but unconventional — move to the top with the other imports.

2. File: src/utils/processUtils.ts -> Line 5: when the child is killed by a signal (e.g. `timeout` fires), `result.status` is `null`, so the message reads "Process exited with status null". Include `result.signal` in the error for debuggability.

3. File: src/cli/setupClaude.ts -> Line 83: `shell: process.platform === "win32"` re-enables a shell on Windows (needed for `.cmd` resolution, so justified) — but that's only injection-safe because the command and args are hardcoded constants. Worth a note that this pattern must not be copied with user-controlled input; also the `timeout` kill applies to `cmd.exe` and may orphan the grandchild process (same as the old `execSync`, so no regression).

The duplication in `setup-claude.js` (inline checks instead of `checkSpawnResult`) is acknowledged in a comment and reasonable given the JS/TS boundary. `validate-hooks.js` changes preserve prior behavior on every edge case (spawn error, timeout, null status). Otherwise, this looks good.

---

giauphan:
cc @jules

I've reviewed this PR — it's a well-executed security hardening change (replacing `execSync` string commands with `spawnSync(process.execPath, [...], { shell: false })`). The `validate-hooks.js` rewrite is behavior-preserving (I checked all five call sites: exit-code checks, timeout handling, and the `echo | bash` → `input` conversion all match original semantics, and passing paths as args also fixes word-splitting on spaces). Only minor nits:

- File: src/utils/processUtils.ts -> Line 7: When a child is killed by a signal, `status` is `null` and the thrown message reads `"Process exited with status null (signal SIGTERM)"` — confusing; consider special-casing `status === null` with `result.signal`.
- File: src/utils/processUtils.ts -> Line 3: Prefer a generic signature `checkSpawnResult<T>(result: SpawnSyncReturns<T>)` over the `string | Buffer` union — matches every call site (`SpawnSyncReturns<Buffer>` from `stdio: "inherit"`) directly instead of relying on TS's unsound covariant property check.
- File: src/cli/commands.ts -> Line 664: The `import { checkSpawnResult }` sits mid-file; ESM hoisting makes it work, but it should be at the top with the other imports.

The `shell: process.platform === "win32"` in setupClaude.ts is acceptable as-is since both command and args are static (needed to launch `.cmd` wrappers post-CVE-2024-27980), and the comment documents it. Everything else looks good.

---

giauphan:
cc @jules

I've reviewed the diff. The `execSync` → `spawnSync` migration is done correctly overall — argv arrays, no shell interpolation, and the new `checkSpawnResult` helper handles the signal/null-status edge case properly. Only minor issues worth noting:

**File: src/cli/setupClaude.ts** -> Line 85: Inline checks re-implement `checkSpawnResult` (which was created in this PR) instead of using it, and lose the signal handling: on `timeout: 5000` the process is killed by signal, `result.status` is `null`, and the error message becomes "Process exited with status null" instead of "Process killed by signal SIGTERM". Suggest replacing lines 84–85 with `checkSpawnResult(result)`.

**File: src/cli/setup-claude.js** -> Line 33: Duplicated spawn/error-check logic that now exists in three places (`processUtils.ts`, `setupClaude.ts`, `setup-claude.js`). The inline comment explains why (JS can't import TS), but if `setup-claude.js` is compiled/bundled alongside the TS sources it could import from the compiled `dist` output instead — worth checking whether one of the two setup files is dead code that can be removed.

Verified non-issues, for the record: stdin piping in test 8 (`input` vs `echo |`) is equivalent for JSON consumption; timeout behavior in `validate-hooks.js` checks correctly fails validation in all four cases; `shell: process.platform === "win32"` in `setupClaude.ts` is the right call for `.cmd` shims with static args; the `processUtils.ts` typing (`SpawnSyncReturns<T>` generic) is correct and infers cleanly at both call sites in `commands.ts`.

No correctness, security, or performance problems found — the PR is a genuine hardening improvement.

---

giauphan:
cc @jules

Overall this PR is a solid security/correctness improvement — replacing `execSync` with `spawnSync` and array args removes shell injection surface and fixes silent quoting bugs with spaces in homedir paths. Behavior in the hook validators is preserved (non-zero exit → `false`, exit 0 → `true`), and `input:` instead of `echo | bash` is cleaner. A few minor things worth fixing:

1. **`src/utils/processUtils.ts` -> Line 9-10**: The `signalMsg` branch is effectively dead code — Node sets `status: null` whenever a process is terminated by a signal, and that case already returns at line 7. Meanwhile `status === null` *without* a signal falls through to the confusing message `"Process exited with status null"`. Simpler and correct: `if (result.status === null) throw new Error(\`Process terminated by ${result.signal ?? "unknown signal"}\`)`, then the plain status throw.

2. **`src/cli/setup-claude.js` -> Line 35**: Same issue — `status` is `null` when killed by a signal, so the error reads "Process exited with status null". Since the comment says this can't import the TS helper, at least mirror the null handling: `result.status === null` → report `result.signal`.

3. **`.jules/sentinel.md` -> Line 39**: New entry is dated `2025-03-01`, which is over a year in the past relative to today (2026-09-12) — looks like a backdated/wrong date.

4. **`scripts/validate-hooks.js` -> Line 118 (bash -n check)**: `result.error` isn't checked, so a spawn failure (e.g., bash missing) is conflated with a syntax error — both return `false`. Acceptable for a validator, but a `result.error` guard would make failures diagnosable.

Non-blocking: in `setupClaude.ts` the Windows `shell: process.platform === "win32"` trick is the right call for `.cmd` shims, and args are static so no injection risk. Everything else (no secrets, no perf regressions, `spawnSync` blocking is fine for CLI scripts) looks good.

---

giauphan:
cc @jules

The PR files aren't in this workspace, so I reviewed the diff on its own merits. Overall this is a clean, well-executed security refactor: `execSync` → `spawnSync` with `shell: false`, `process.execPath` pinning, and the new helper correctly orders the `error` / `status === null` / `signal` checks. I traced each conversion and the semantics are preserved (a few spots got slightly stricter, which is fine — e.g. requiring exit 0 in the brain-context check matches the old `execSync` throw behavior). It also fixes a latent bug: `bash ${wrapperPath}` under a shell broke on home directories containing spaces; argv-style `spawnSync` handles that.

Only minor issues:

- File: src/cli/setup-claude.js -> Line 33: Inline status/error checks duplicate `checkSpawnResult` from `src/utils/processUtils.ts`. The comment justifies it (plain-JS script can't import the TS util), but a small plain-`.js` helper would avoid a second copy of this logic drifting.
- File: .jules/sentinel.md -> Line 39: The new learning entry has garbled phrasing ("Relying on the string `'node'` is unsafe if the environment path is attacker-controlled" is written as "Relaying..."). Hurts readability of the recorded lesson.
- File: scripts/validate-hooks.js -> Line 14: The `execSync` import was removed; verify the file has no other `execSync` references outside the hunks shown, or it would throw a ReferenceError at those call sites (can't confirm from the diff alone).

No correctness, security, or performance problems introduced — the `shell: process.platform === "win32"` in setupClaude.ts is the right call for npm `.cmd` shims, and the static `--version` arg keeps it injection-safe. Everything else looks good.

---

giauphan:
cc @jules

The target repo isn't checked out here, so this is a standalone review of the diff. Overall this is a clean, correct security hardening — I traced each `execSync` → `spawnSync` conversion and behavior is preserved in all four `validate-hooks.js` tests (test 7's exit-code check matches the old `execSync` throw semantics; test 8's `input:` correctly replaces the `echo |` pipe). No correctness, security, or performance issues found.

Minor items:

- File: src/cli/setup-claude.js -> Line 34: Inline error/status checks duplicate `checkSpawnResult`. The comment explains why (JS can't import TS), but the copies can drift — a small plain-JS shared helper would avoid this if more JS scripts need it. Acceptable as-is for one call site.
- File: src/cli/setupClaude.ts -> Line 85: On failure the new error ("Process exited with status 1") drops the child's stderr that `execSync`'s thrown error used to include, making the warn message less diagnostic. Consider appending `result.stderr?.trim()` to the thrown error.
- File: src/utils/processUtils.ts -> Line 4: Throwing the raw `result.error` (e.g., ENOENT) loses which command failed. Optional: wrap with the command/args for context since this helper is generic.

One completeness note: the catch in `setupClaude.ts` mentions an "npx/local path" fallback — verify no `execSync` call sites remain elsewhere in the codebase (e.g., that fallback), since the PR only converts the ones shown in the diff.

---

PR #225, opened 2026-09-14
🛡️ Sentinel: [MEDIUM] Fix TOCTOU vulnerability in appendFileSync

giauphan:
🚨 **Severity**: MEDIUM
💡 **Vulnerability**: The `export_team_artifact` tool used `fs.appendFileSync` to modify `.gitignore`. `fs.appendFileSync` naturally follows symlinks. An attacker could potentially replace the `.gitignore` with a symlink pointing to a sensitive system file, tricking the server into appending data to it (a TOCTOU vulnerability).
🎯 **Impact**: Arbitrary file append to files accessible by the process, potentially allowing for minor corruption or bypass depending on the target file.
🔧 **Fix**: Implemented `appendFileSyncNoFollow` using `fs.openSync` with `O_NOFOLLOW` and used it instead of `fs.appendFileSync`.
✅ **Verification**: Verified using `git diff`, and running `pnpm lint`, `pnpm build`, `pnpm test`.

Additionally, documented the vulnerability and fix in `.jules/sentinel.md`.

---
*PR created automatically by Jules for task [8738438291042178434](https://jules.google.com/task/8738438291042178434) started by @giauphan*


giauphan:
cc @jules

The target repo isn't in this workspace, so I'm reviewing the diff on its own. The fix itself is sound — here's what I found:

**Verdict:** The core approach is correct: `O_CREAT | O_WRONLY | O_APPEND | O_NOFOLLOW` + fd-based write + `closeSync` in `finally` is exactly the right pattern, and `fs.appendFileSync(fd, ...)` with a descriptor is valid Node API usage (fd is not closed by it, hence the `finally`). Two issues worth attention:

src/utils/pathUtils.ts -> Line 91: `fs.constants.O_NOFOLLOW` is `undefined` on Windows (it's POSIX-only), so the bitwise OR silently coerces it to `0` and the function follows symlinks — the exact guarantee this helper exists for quietly disappears on win32, with no error. Throw (or at least log) when the constant is unavailable instead of degrading silently. Note `writeFileSyncNoFollow` likely has the same latent issue; this PR extends the pattern to a second call site.

src/presentation/mcpServer.ts -> Line 3334: a user who legitimately symlinks `.gitignore` (not uncommon) will now get an `ELOOP` failure instead of a silent append. That's the intended trade-off, but make sure the surrounding `export_team_artifact` code catches this and degrades to a warning rather than failing the whole tool call.

Minor:

src/utils/pathUtils.ts -> Line 85: default `mode: 0o600` means a newly created `.gitignore` gets stricter perms than the `0644` a plain append would produce — mostly moot given the `existsSync` guard, but worth confirming it matches `writeFileSyncNoFollow`'s default for consistency in a "shared with team" flow.

Everything else looks good — no logic bugs in the fd handling, no fd leaks (`finally` covers throws), no injection surface, and the sentinel.md documentation accurately describes the vulnerability and fix.

---

giauphan:
Superseded by PR #235.

PR #229, opened 2026-09-17
🛡️ Sentinel: [CRITICAL] Fix TOCTOU Symlink Write Vulnerability in gitignore append

giauphan:
🛡️ Sentinel: [CRITICAL] Fix TOCTOU Symlink Write Vulnerability in gitignore append

🚨 Severity: CRITICAL
💡 Vulnerability: The `export_team_artifact` tool used `fs.appendFileSync` to append `.codeatlas` configuration to `.gitignore`. Because `fs.appendFileSync` follows symlinks, an attacker could place a symlink at `.gitignore` to arbitrarily overwrite or append to any file on the system, creating a TOCTOU (Time-of-Check to Time-of-Use) vulnerability.
🎯 Impact: An attacker could potentially escalate privileges, modify sensitive configuration, or append backdoors by exploiting the predictable `.gitignore` write path if they control the workspace.
🔧 Fix: Created `appendFileSyncNoFollow` in `src/utils/pathUtils.ts` which uses `fs.openSync` with `fs.constants.O_NOFOLLOW` flag to prevent following symlinks. Replaced the insecure `fs.appendFileSync` call in `src/presentation/mcpServer.ts` with this secure alternative.
✅ Verification: Ran `pnpm build && pnpm test` to verify changes, test suite passes. Also verified function behavior with a manual script that appending works and `O_NOFOLLOW` is passed.

---
*PR created automatically by Jules for task [16294907526982941508](https://jules.google.com/task/16294907526982941508) started by @giauphan*


giauphan:
cc @jules

Review of the diff (note: the changed files aren't in this workspace — this is the github-review-bot repo, the PR touches a different codebase — so this is a static review of the diff alone).

Overall the change is correct: `appendFileSyncNoFollow` is the right pattern (`openSync` with `O_CREAT|O_WRONLY|O_APPEND|O_NOFOLLOW`, write, close in `finally`), and it matches the mitigation described in the sentinel note. A few things worth flagging:

- File: src/presentation/mcpServer.ts -> Line 3334: Uncaught `ELOOP` on the exact attack path. If `.gitignore` is a symlink (the scenario this fix defends against), `openSync` with `O_NOFOLLOW` now throws instead of silently writing. The call site has no try/catch, so a malicious symlink turns a silent write into an abrupt tool-call failure. Catch the error here (or inside the helper) and log/skip gracefully — failing closed is right, failing noisy is not.
- File: src/utils/pathUtils.ts -> Line 92: `fs.constants.O_NOFOLLOW` is undefined on Windows, so the bitwise OR silently degrades to no symlink protection there (no crash, just no defense). Acceptable if the pre-existing `writeFileSyncNoFollow` has the same behavior, but worth a one-line comment or a guard for consistency.
- File: src/utils/pathUtils.ts -> Line 85: Minor: on creation the helper forces mode `0o600` while an existing `.gitignore` keeps its own perms — fine, just noting the asymmetry with typical `0644` gitignore files in case other tools read it under a restricted account.
- File: .jules/sentinel.md -> Line 43: Date says "2024-03-12" while the surrounding entries/project timeline are 2026 — likely a typo.

Non-issues I checked and ruled out: `fs.appendFileSync(fd, ...)` with a file descriptor is valid and uses the `O_APPEND`-opened fd correctly; the `finally`/`closeSync` is correct on the error path; the remaining `existsSync → read → append` race can only cause duplicate gitignore entries (pre-existing, benign).

The fix itself is sound — just add ELOOP handling at the call site.

---

giauphan:
cc @jules

Here's my review of the diff:

## Overall: Solid security improvement. No blocking issues.

### Issues that matter:

**1. `sentinel.md` — "TOCTOU" claim is overstated**

The new entry says symlink following prevents TOCTOU attacks using `O_NOFOLLOW | O_APPEND`. This isn't technically correct. `O_NOFOLLOW` only prevents symlink-following at the moment `openSync` is called — it doesn't prevent TOCTOU (time-of-check-to-time-of-use). The vulnerability here is simpler: `fs.appendFileSync` blindly follows symlinks regardless. The real protection is just symlink rejection, not TOCTOU mitigation. Suggest rewording for accuracy.

**2. `writeFileSyncNoFollow` — harmless but noisy refactor**

```js
const flags = fs.constants.O_CREAT | fs.constants.O_WRONLY | fs.constants.O_TRUNC | (fs.constants.O_NOFOLLOW || 0);
```

The `(O_NOFOLLOW || 0)` fallback is defensive against environments where `fs.constants.O_NOFOLLOW` might be undefined, but this constant has been present since Node.js 0.x on non-Windows platforms and is always available on Linux/macOS. It makes the code slightly less readable without meaningful safety gain. Either use `fs.constants.O_NOFOLLOW` directly, or document why the fallback exists.

### Good things:

- **New `appendFileSyncNoFollow`** — clean implementation, mirrors `writeFileSyncNoFollow` correctly, uses the fd-based pattern properly. Note: inside the try block it calls `fs.appendFileSync(fd, content)` which works because the fd position is maintained by `O_APPEND`; the function name is slightly confusing but functionally correct.

- **ELOOP handling in mcpServer.ts** — gracefully ignores malicious symlinks with a warning rather than crashing the export flow. The catch narrows `err` to `NodeJS.ErrnoException` and checks `error.code === 'ELOOP'`, which is the right error code for symlink loops/rejection.

- **Documentation update** — good to have in `sentinel.md` for future reference.

### Minor notes (not blockers):

- The ELOOP case logs via `console.warn` — consider adding the file path to the message for debugging purposes, e.g., `console.warn("[Export Artifact] Ignored malicious symlink at %s", gitignorePath)`.
- No TypeScript type issue (`err as NodeJS.ErrnoException`) aligns with the project convention of avoiding strict casts when possible, though `err instanceof Error && 'code' in err` would be more idiomatic TS.

---

giauphan:
cc @jules

Overall this is a solid security fix — the symlink-following TOCTOU on `.gitignore` is correctly addressed, and error handling is sane. A few minor notes, nothing blocking:

**File: src/presentation/mcpServer.ts** -> Line 3338: The warning "Ignored malicious symlink" also fires for *legitimate* symlinked `.gitignore` files (e.g., a shared template symlink, a common pattern), and the function still reports success even though the entry was never written. Consider wording like "skipped symlinked .gitignore" and including the skip in the tool's result output so callers know config wasn't persisted.

**File: src/presentation/mcpServer.ts** -> Line 3331: The `existsSync`/`readFileSync` guard is still TOCTOU-racy (file could be swapped for a symlink between read and append), but since `appendFileSyncNoFollow` re-validates at open time, it's harmless — just noting the outer check is advisory only.

**File: src/utils/pathUtils.ts** -> Line 80: `(fs.constants.O_NOFOLLOW || 0)` silently degrades to following symlinks on platforms without `O_NOFOLLOW` (Windows). That's reasonable graceful degradation, but it's a silent loss of protection — worth a one-line comment, and both functions now duplicate the flag-building logic; a tiny shared `safeFlags(extra)` helper would keep them in sync.

**File: src/utils/pathUtils.ts** -> Line 79: Default mode `0o644` here vs `0o600` for `writeFileSyncNoFollow`. Correct for `.gitignore`, but if this helper gets reused for sensitive files the permissive default is a footgun — fine as-is, just be aware.

Positives: `try/finally` guarantees the fd is closed, ELOOP is the right errno for `O_NOFOLLOW` (other errors correctly re-thrown), the `.jules/sentinel.md` documentation matches the implementation, and no `any` types or secrets introduced. No correctness or performance concerns.

---

giauphan:
cc @jules

I've reviewed the diff — let me walk through what I see.

This PR adds `appendFileSyncNoFollow` to securely append to files without following symlinks, mirrors the existing `writeFileSyncNoFollow` pattern. It updates the `.gitignore` append in `export_team_artifact` to use the new function with ELOOP error handling, and documents the vulnerability in sentinel.md.

**Issues I found:**

`src/utils/pathUtils.ts` -> Line 74: The cast `(fs.constants.O_NOFOLLOW || 0)` is harmless but confusing — `O_NOFOLLOW` is always either a number or `undefined` in Node.js builds, so the `|| 0` workaround for non-symlink platforms works but the intent isn't clear. Consider a comment like `// O_NOFOLLOW not supported on all platforms`.

`src/presentation/mcpServer.ts` -> Line 3337: Variable naming inconsistency inside the catch block — `const error = err as ...` followed by `error.code` but then `throw err` (not `throw error`). This is functionally correct but inconsistent with the established convention. Using `throw error` would be cleaner.

`src/utils/pathUtils.ts` -> Line 85: Uses `fs.appendFileSync(fd, content)` on an already-open fd. Node.js supports this (it appends to the fd), but using `fs.writeFileSync(fd, content)` would actually be simpler since there's nothing else to read concurrently — same underlying mechanism, fewer cognitive overhead. Not a bug, just a suggestion.

**Overall: LGTM.** The core fix is correct — `O_NOFOLLOW` + `O_APPEND` via `fs.openSync` properly prevents the TOCTOU/symlink attack that `fs.appendFileSync` is vulnerable to. The sentinel doc and test are appropriate.

---

giauphan:
cc @jules

I've analyzed this PR diff. Let me break down what I found.

## Review

This PR fixes a symlink write vulnerability by adding `appendFileSyncNoFollow` and updating `export_team_artifact` to use it. Here are my findings:

### Issues

**File: src/utils/pathUtils.ts -> Line 75**: `(fs.constants.O_NOFOLLOW || 0)` is unnecessary — bitwise OR already coerces `undefined` to `0` (`undefined | 0 === 0`). Same on line 86. You can remove the `|| 0` shorthand to keep things DRY, or leave it as-is for explicit defensive coding.

**File: src/presentation/mcpServer.ts -> Line 3335**: Graceful degradation only on `'ELOOP'` — other symlink-related errors like `EPERM` are re-thrown via `throw error`. Consider whether you also want to catch those, or if letting them bubble is intentional.

**File: src/presentation/mcpServer.ts -> Line 3334**: Only updates `.gitignore` if it already exists. If a project has no `.gitignore`, the `.codeatlas/` exception won't be added at all. This may be intentional (don't create user files), but worth noting.

### What's Good

- **Security**: The core fix is solid. Using `fs.openSync` with `O_NOFOLLOW` + `O_APPEND` prevents symlink-based TOCTOU attacks, matching the sentinel doc entry.
- **Resource management**: The rewritten `writeFileSyncNoFollow` now uses `try/finally` with `closeSync` — the old version left the fd unclosed.
- **Error handling**: Catching `'ELOOP'` specifically and converting to a warning + note message is a clean pattern.
- **Cross-platform awareness**: Comments acknowledging `O_NOFOLLOW` not supported on Windows are appropriate.
- **Types**: No `any` types, proper `NodeJS.ErrnoException` typing.

---

giauphan:
Superseded by PR #235.

PR #231, opened 2026-09-19
⚡ Bolt: Optimize O(L) relationship tracing generation with early exit

giauphan:
💡 What: Replaced the inefficient `.filter().slice().map()` chain used when collecting relationships in `trace_relationships` with a single `for...of` loop over `links` that has an early exit.
🎯 Why: In large graph datasets with hundreds of thousands of links, calling `.filter()` forces a full O(L) traversal of the entire array, followed by two separate intermediate array allocations for `slice` and `map`, even though only 50 elements were requested.
📊 Impact: Avoids allocating massive intermediate `traceLinks` arrays and reduces time complexity from processing all O(L) links (e.g. 200,000+ items) down to early-breaking after matching the first 50 results. Saves significant CPU and GC overhead per query.
🔬 Measurement: Run the test suite and observe successful compilation and equivalent correctness results.

Also added the learning to `.jules/bolt.md`.

---
*PR created automatically by Jules for task [13684989861888750913](https://jules.google.com/task/13684989861888750913) started by @giauphan*


giauphan:
cc @jules

Overall the change is a sound optimization — the early exit preserves the original `slice(0, 50)` semantics since `traceLinks` was only used for `relationships`, and matched links keep the same order. Two minor flags:

- File: src/presentation/mcpServer.ts -> Line 1164: `const relationships = [];` uses implicit `any[]` inference, violating the project's "avoid any types" rule. Annotate it, e.g. `const relationships: Array<{ from: string; to: string; type: string }> = [];` (or match the actual link `type` union).

- File: src/presentation/mcpServer.ts -> Line 1163: the "⚡ Bolt Optimization" comment adds no information beyond what the code already says — consider trimming it to nothing or a one-liner. The magic `50` is also duplicated from the old `.slice(0, 50)`; a named constant (e.g. `MAX_RELATIONSHIPS`) would prevent the two from drifting apart.

Otherwise correct: worst case remains O(L) when fewer than 50 matches exist (same as before), typical case is strictly better, and the `from`/`to` label fallback logic is unchanged.

---

giauphan:
cc @jules

Updated 1.56.2 → 1.58.0
The file isn't in this workspace (the PR targets the downstream codebase), so I reviewed it from the diff.

**Review result: looks good — no issues that matter.**

- **Correctness:** The new `for...of` loop is an exact behavioral replacement for `links.filter(...).slice(0, 50).map(...)`: same predicate (`visited.has` on both endpoints), same result order (insertion order of `links`), same cap of 50. No edge cases lost.
- **Performance:** Improvement is real — besides the early exit, it also eliminates the intermediate `traceLinks` array, which previously allocated one entry per matched link across the whole graph even though only 50 were used.
- **Security:** No changes to input handling, no new secrets or unsafe patterns.
- **Code quality / Maintainability:** Clean and consistent with the existing loop-based patterns elsewhere in the file (e.g., the `byFile` loop right below). The `.jules/bolt.md` learning-log entry accurately describes the change.

Only trivial nit (not worth blocking): `MAX_RELATIONSHIPS` could live at module scope if this pattern recurs, but inline is fine here. Also double-check `traceLinks` isn't referenced anywhere outside the diff hunks — based on the diff it had exactly one definition and one usage, both removed.

---

giauphan:
Duplicate of PR #228.

